Issue #3, opened 2018-10-12
Creating an FAT32 block file


Eun:
I am trying to create a FAT32 block file with files in it.
Basicly I want to achive the same as:
```bash
dd if=/dev/zero of=test.raw bs=100M count=10
mount test.raw /tmp/fs/
echo Hello World > /tmp/fs/test
umount /tmp/fs/
```

I made my way upto creating the file but I was unable to write some data on it.
(The functions succeed but after mounting the file nothing is on the partition)

Any advice?

```go
package main

import (
    "os"
    "testing"

    "github.com/deitch/diskfs/disk"
    "github.com/deitch/diskfs/filesystem"
)

func TestFat32(t *testing.T) {
    f, err := os.Create("test.raw")
    if err != nil {
        panic(err)
    }
    // make it a 10MB file
    f.Truncate(10 * 1024 * 1024)

    fileInfo, err := f.Stat()
    if err != nil {
        t.Fatalf("Error reading info on temporary disk: %v", err)
    }

    d := &disk.Disk{
        File: f,
        LogicalBlocksize: 512,
        PhysicalBlocksize: 512,
        Info: fileInfo,
        Size: fileInfo.Size(),
    }
    // // this is partition start and end in sectors, not bytes
    // sectorSize := 512
    // partitionStart := uint64(2048)
    // // make it a 5MB partition
    // partitionSize := uint64(5 * 1024 * 1024 / sectorSize)
    // partitionEnd := partitionSize + partitionStart - 1

    // // apply the partition table
    // err = d.Partition(&gpt.Table{
    //  Partitions: []*gpt.Partition{
    //      {
    //          Start: partitionStart,
    //          End: partitionEnd,
    //          Type: gpt.EFISystemPartition,
    //          Name: "Disk",
    //      },
    //  },
    //  LogicalSectorSize: sectorSize,
    // })
    // if err != nil {
    //  panic(err)
    // }

    fs, err := d.CreateFilesystem(0, filesystem.TypeFat32)
    if err != nil {
        panic(err)
    }
    rw, err := fs.OpenFile("/test", os.O_CREATE|os.O_RDWR)
    if err != nil {
        panic(err)
    }

    _, err = rw.Write([]byte("Hello World"))
    if err != nil {
        panic(err)
    }
}
```

deitch:
You really shouldn't create the `Disk` struct manually, as you miss the critical initialization. You should do it via `diskfs.Create()`:

```go
[...]

PR #5, opened 2018-10-18
Fix missing directory entry write for new file

deitch:
When creating a new file, it was creating the directory entry in the parent, but not actually writing the parent out. So the data was on disk, but the file pointer was not there.

@Eun can you take a look at this branch please? Not complete, but good to get feedback.

For #3


deitch:
I notice there still is an issue with `Mkdir()` in fat32 where it doesn't always show up. That should be a separate issue.

Eun:
LGTM :tada: Thanks :+1:

PR #11, opened 2019-03-21
Fix test for GetPartitionSize, and fix GetPartitionSize for GPT

### LINKED ISSUE #10: gpt Table.GetPartitionSize() returning incorrect size
The Size field in gpt.Partition is in bytes, but gpt.Table.GetPartitionSize() is multiplying the  gpt.Partition.Size by the sector size (see https://github.com/diskfs/go-diskfs/blob/master/partition/mbr/table.go#L181)

This caused disk.CreateFilesystem() to fail on a 4GiB gpt partition because it said the size was > max FAT32 filesystem size.

deitch:
Fixes #10 

Two issues:

1. Test for gpt `GetPartitionSize()` was comparing sizes of partitions that have no size at all, so 0*anything still was 0, and test passed, even though it should have failed. Test now checks first partition, which has proper size.
2. gpt `GetPartitionSize()` was multiplying partition size by sector size, even though size already was in bytes.

### DIFF
--- a/partition/gpt/table.go
+++ b/partition/gpt/table.go
@@ -518,7 +518,8 @@ func (t *Table) GetPartitionSize(partition int) (int64, error) {
 		return 0, fmt.Errorf("Requested partition %d but only have %d partitions in table", partition, len(t.Partitions))
 	}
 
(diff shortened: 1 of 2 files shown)


queuem:
Patch looks good. Thanks for fixing so quickly!

PR #18, opened 2019-08-05
MBR partition writeContents() was not checking against the number of …

a2geek:
…bytes in the partition when checking size. Replicated logic from readContents().


deitch:
No arguments on this, other than replicated logic (which isn't DRY, but sometimes is unavoidable). Let's hold on this until #19 is updated and merged in, then this can be rebased and it will be a much simpler PR.

PR #19, opened 2019-08-06
filesystem labels

a2geek:
I needed labels for a project I'm working on. They are recognized by Linux, so everything should be good from that perspective. This specific implementation adds a Disk#CreateFilesystemSpecial that takes a structure with parameters... with the assumption that additional details can just be added to the structure without having to then change signatures all around.

Sorry, the MBR write came along with the PR... I had committed and pushed before I realized I needed a branch!

### DIFF
--- a/partition/mbr/partiton.go
+++ b/partition/mbr/partiton.go
@@ -111,8 +111,6 @@ func (p *Partition) writeContents(f util.File, logicalSectorSize, physicalSector
 	b := make([]byte, physicalSectorSize, physicalSectorSize)
 	// we start at the correct byte location
 	start := p.Start * uint32(logicalSectorSize)
-	size := p.Size * uint32(logicalSectorSize)
-
 	// loop in physical sector sizes
 	for {
 		read, err := contents.Read(b)
@@ -120,7 +118,7 @@ func (p *Partition) writeContents(f util.File, logicalSectorSize, physicalSector
 			return total, fmt.Errorf("Could not read contents to pass to partition: %v", err)
 		}
 		tmpTotal := uint64(read) + total
-		if uint32(tmpTotal) > size {
+		if uint32(tmpTotal) > p.Size {
 			return total, fmt.Errorf("Requested to write at least %d bytes to partition but maximum size is %d", tmpTotal, p.Size)
 		}
 		var written int
@@ -138,7 +136,7 @@ func (p *Partition) writeContents(f util.File, logicalSectorSize, physicalSector
 		}
 	}
 	// did the total written equal the size of the partition?
-	if total != uint64(size) {
+	if total != uint64(p.Size) {
 		return total, fmt.Errorf("Write %d bytes to partition but actual size is %d", total, p.Size)
 	}
 	return total, nil
--- a/partition/mbr/partiton.go
+++ b/partition/mbr/partiton.go
@@ -111,6 +111,8 @@ func (p *Partition) writeContents(f util.File, logicalSectorSize, physicalSector
 	b := make([]byte, physicalSectorSize, physicalSectorSize)
 	// we start at the correct byte location
 	start := p.Start * uint32(logicalSectorSize)
+	size := p.Size * uint32(logicalSectorSize)
+
 	// loop in physical sector sizes
 	for {
 		read, err := contents.Read(b)
@@ -118,7 +120,7 @@ func (p *Partition) writeContents(f util.File, logicalSectorSize, physicalSector
 			return total, fmt.Errorf("Could not read contents to pass to partition: %v", err)
 		}
 		tmpTotal := uint64(read) + total
-		if uint32(tmpTotal) > p.Size {
+		if uint32(tmpTotal) > size {
 			return total, fmt.Errorf("Requested to write at least %d bytes to partition but maximum size is %d", tmpTotal, p.Size)
 		}
 		var written int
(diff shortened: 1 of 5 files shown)


deitch:
So basically having the ability to pass a `FilesystemSpec struct` to create a filesystem, so it can have more options? Why bother with `CreateFilesystemSpecial()`? Why not just replace `CreateFilesystem()` that takes two params with the `struct`? I see no need to keep both of them. With go modules and hashes, we don't need to worry as much about backwards compatibility.

a2geek:
Certainly. Go's still a 2nd language to me, so I was trying to avoid conflicts with existing code... but, if that's not an issue, it certainly can be switched around! 

The concept on the struct is just to enable more "features" to be added in the future as well as to have "optional" stuff that doesn't have to be included in the list of parameters.

deitch:
Definitely you’re on the right approach. May need to update the tests as well, and the README, but those are easy concerns.

a2geek:
I've updated the `CreateFilesystem()` and dropped the `CreateFilesystemSpecial()` functions. The tests have also been updated. I did change the `fat32_test` package to `fat32` since most of the tests were already in the `fat32` package.

Unfortunately, I am unable to run the unit tests but I can't quite get why. Here's what I see:
```
$ make unit-test
<snip>
# github.com/a2geek/go-diskfs/filesystem/fat32 [github.com/a2geek/go-diskfs/filesystem/fat32.test]
filesystem/fat32/fat32_test.go:53:28: undefined: fat32.Fat32File
filesystem/fat32/fat32_test.go:55:63: undefined: fat32.Fat32File
filesystem/fat32/fat32_test.go:70:71: undefined: fat32.Fat32File
filesystem/fat32/fat32_test.go:73:85: undefined: fat32.Fat32File
<snip>
```
Note that I get a bunch of undefined's and I don't understand why since *I* can find them!!
* `fat32.Fat32File`
* `iso9660.GetTestFile`
* `iso9660.ISO9660File`
* `iso9660.RockRidgeFile`
* `mbr.GetValidTable`

I'm running `go version go1.12.6 linux/amd64`. And I pulled in the updated `go.mod` from the latest commit since that's what I'm vaguely familiar with and `dep` didn't seem to be working for me. Just in case any of that matters.

deitch:
> I did change the fat32_test package to fat32 since most of the tests were already in the fat32 package.

I set these up as `package fat32_test` instead of `package fat32` in order to keep as many tests as possible _without_ access to non-exported funcs/structs/vars/etc. It is generally considered very good practice in go, in order to properly test the API. I use `package fat32` in a test file only inside a file named `fat32_internal_test.go` to distinguish those, and only when I am actively unit testing a non-exposed function. For example:

```go
package foo

func Triple(d int) int {
  return multiply(d, 3)
}
func Double(d int) int {
  return multiply(d, 2)
}

func multiple(d, factor int) int {
  return d*factor
}
```

In the above, I am exporting `Triple()` and `Double()`, but not `multiple()`, so I would test `Triple()` and `Double()` in a file named `foo_test.go` in the package `foo_test`, but `multiply()` in a file named `foo_internal_test.go` in the package `foo`.

This isn't any kind of official standard or anything, and not everyone agrees with it, but I like the encouraged separation it provides.

deitch:
> Note that I get a bunch of undefined's and I don't understand why since I can find them!!

Almost certainly the package renaming. Switch it back, see what happens.

partition/mbr/partiton.go
```diff
@@ -111,14 +111,16 @@ func (p *Partition) writeContents(f util.File, logicalSectorSize, physicalSector
 	b := make([]byte, physicalSectorSize, physicalSectorSize)
 	// we start at the correct byte location
 	start := p.Start * uint32(logicalSectorSize)
+	size := p.Size * uint32(logicalSectorSize)
```
deitch: Why these changes too MBR?

a2geek:
Ok, renamed the package back. Still no luck. But, good news is that I checked out the source into a separate directory, and I did confirm the unit tests work.

My guess at the moment is because I am doing a replace with `github.com/diskfs/go-diskfs` to my fork.  After a `go tidy`, it ends up looking like this:
```
module go-diskfs

go 1.12

require (
	github.com/diskfs/go-diskfs v0.0.0-00010101000000-000000000000
	github.com/kr/pretty v0.1.0 // indirect
	github.com/satori/go.uuid v1.2.0
	golang.org/x/sys v0.0.0-20190804053845-51ab0e2deafa
	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
)

replace github.com/diskfs/go-diskfs => github.com/a2geek/go-diskfs v0.0.0-20190810185729-cb26ff0fd076
```
... and I've also tried removing the stuff that's not in the original `go.mod`. (I swapped over to go modules since I was having issues with dep and am somewhat familiar with Go modules.)

partition/mbr/partiton.go
a2geek: Let me pull that out. I didn't realize I was making multiple changes, so didn't fork originally and mucked up the commit when I did. These changes are really part of PR #18.

deitch:
Hmm, I suggest saving your changes (they only are a few lines in a few files), blow away your branch, rebase and then `checkout -b ` again and copy your changes in.

I moved this over to go modules last week so a rebase would help.

a2geek:
I did that and a bit more (likely the least efficient route!). I think this PR is at a better state now -- unit tests to pass on my machine.

deitch:
Looks great to me. tests are clean. Thank you for this!

PR #23, opened 2019-08-15
Replace illegal characters in shortname with _

ibuildthecloud:
When calculating the 8.3 filename we replace all characters
that are not [A-Z0-9_] with _

Signed-off-by: Darren Shepherd <darren@rancher.com>

### DIFF
--- a/filesystem/iso9660/finalize.go
+++ b/filesystem/iso9660/finalize.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
We are attempting to use this library to create cloud-init ISO files.  A cloud-init ISO has two files in it: `user-data` and `meta-data`.  When creating the ISO it would fail validating the shortname `META-DATA;1`.  This PR will replace all illegal characters in the shortname with `_` so `meta-data` becomes `META_DATA;1`.  This behavior seems conesistent with the default behavior of `genisoimage` (although, `genisoimage` will truncate to `META_DAT.;1` also.

deitch:
Ah, so basically cleaning it up at the shortname/extension phase during `walkTree`, so that the illegal name doesn't even get passed? That makes a lot of sense.

> although, genisoimage will truncate to META_DAT.;1 also

I would need to reread the spec, but I am pretty sure that the 8.3 shortname is supposed to be 8 chars, then the period, then the 3, with the period not counting as part of the 8. So it probably should be `META_DAT;1` and not `META_DATA;1`. Will recheck the spec. Other than that, looks fine and thank you.

deitch:
OK, yes, in ecma-119, at L1:

> a File Name shall not contain more than eight d-characters or eight d1-characters

So `meta-data` should become `META_DAT` if correct (or `META_DAT;1` if including the proper version number). The question is why we didn't truncate it to begin with. We actually do not truncate it anywhere. 

Again, the vast majority of readers handle the violations just fine, but trying to stay spec compliant. Probably should have other options here.

In the meantime, this looks clean, and the `meta-data` to `META_DATA` instead of to `META_DAT` should be treated as a separate issue.

PR #24, opened 2019-08-25
Allowing MBR writeContents() to write successfully

a2geek:
MBR partition writeContents() was not checking against the number of bytes in the partition when checking size. Replicated logic from readContents() and updated unit tests.

Review carefully please as I needed to change more in the unit tests than I would have expected.

### DIFF
--- a/partition/mbr/partition_internal_test.go
+++ b/partition/mbr/partition_internal_test.go
@@ -277,7 +277,7 @@ func TestWriteContents(t *testing.T) {
 			EndHead:       0,
 			EndSector:     2,
 			EndCylinder:   0,
-			Start:         partitionStart,
+			Start:         partitionStart / 512,
 			Size:          1,
 		}
 		// make a byte array that is too big
@@ -290,9 +290,9 @@ func TestWriteContents(t *testing.T) {
 				return len(b), nil
 			},
 		}
-		read, err := partition.writeContents(f, 512, 512, reader)
-		if read != 0 {
-			t.Errorf("Returned %d bytes read instead of 0", read)
+		written, err := partition.writeContents(f, 512, 512, reader)
+		if written != 512 {
+			t.Errorf("Returned %d bytes written instead of 512", written)
 		}
 		if err == nil {
 			t.Errorf("Returned nil error instead of actual errors")
@@ -305,6 +305,7 @@ func TestWriteContents(t *testing.T) {
 
 	t.Run("successful write", func(t *testing.T) {
 		size := 512000
+		sectorSize := size / 512
 		partition := Partition{
 			Bootable:      false,
 			StartHead:     0,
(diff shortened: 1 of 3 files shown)


deitch:
This makes a lot of sense. What I don't understand is why tests didn't fail beforehand?

Also, it looks like more of the tests need to be updated; is that correct?

deitch:
Now I get it. The tests _did_ succeed in catching failures. We got "lucky", because it has a sector size of `1`, and we put larger. Given that, though, your proposed tests are off. I will comment inline.

partition/mbr/partition_internal_test.go
```diff
@@ -277,7 +277,7 @@ func TestWriteContents(t *testing.T) {
 			EndHead:       0,
 			EndSector:     2,
 			EndCylinder:   0,
-			Start:         partitionStart,
+			Start:         partitionStart / 512,
```
deitch: Given that `partitionStart` is only ever used is an entry in `Partition.Start`, and in all cases it should be sector and not bytes (i.e. `/512`), it makes more sense to just change the definition of `partitionStart` from 

```go
const (
        partitionStart   = 2048
)
```

to

```go
const (
        partitionStart   = 2048 / 512
)
```

and then keep `partitionStart` unchanged everywhere it is consumed.

partition/mbr/partition_internal_test.go
```diff
@@ -290,9 +290,9 @@ func TestWriteContents(t *testing.T) {
 				return len(b), nil
 			},
 		}
-		read, err := partition.writeContents(f, 512, 512, reader)
-		if read != 0 {
-			t.Errorf("Returned %d bytes read instead of 0", read)
+		written, err := partition.writeContents(f, 512, 512, reader)
+		if written != 512 {
+			t.Errorf("Returned %d bytes written instead of 512", written)
```
deitch: This is good. We should add a comment that explains why this is, e.g.

```go
// we have a size of 1 sector, or 512 bytes, but are trying to write 2*512. It should write the first and fail on the second
// so we expect an error, along with 512 bytes successfully written
```

partition/mbr/partition_internal_test.go
a2geek: I started down the track, and have a number of test failures. The first one I've figured out suggests that 
```
const (
	mbrPartitionFile = "./testdata/mbr_partition.dat"
)
``` 
needs to be recreated, if that makes sense?

deitch: Should it need to be recreated? We are using the same parameters.

deitch: Hmm, I see them too, but not sure why. Thinking...

deitch:
Actually, revisiting this @a2geek , it looks like the `mbr.img` and the `partitionStart` and `partitionSize` _are_ correct and _are_ in sectors (not bytes). That means that:

* your corrections to `partition.go` are correct
* your corrections to `table_test.go` are correct
* your corrections to `partition_internal_test.go` are partially correct

Will comment inline and resolve the existing ones.

partition/mbr/partition_internal_test.go
deitch: This appears to be incorrect. The `partitionStart` already is in sectors.

a2geek:
Finally! The unit tests are clean now.

deitch:
Looks great. Thanks @a2geek !

PR #27, opened 2019-10-12
Times pkg

luthermonson:
@deitch totally might not know what im doing on this one but felt like moving to a fs times library helps architecture support as the times_<arch> files weren't complete. I was having trouble compiling on other architectures so I switched to this library and was happy with the results. Feel free to check their support matrix: https://github.com/djherbis/times#supported-times

Hope you agree to the change but if you want something different let me know.


deitch:
Hi @luthermonson, and thanks for this.

I have no objection; more than happy to see it expand into new architectures and OSes. I looked at this or some similar library when first doing it, but was concerned about odd dependencies.

I am going to add in GitHub actions to this, so we can get clean checks.

What architectures did you have issues with?

deitch:
OK, GitHub actions enabled. Please rebase on it so we get output.

luthermonson:
@deitch windows and darwin were the ones that didn't like me when i tried to do it. I'll rebase now

deitch:
Thanks. Once it rebases and CI is clean, happy to take it.

luthermonson:
looks like it's failing against a symlink? any advice how to fix?

deitch:
What happens when you run `make test` locally?

luthermonson:
@deitch its failing locally too. ill work against the test and try and give you a new PR

luthermonson:
@deitch you may not want this but changing to Lstat on both the os and times makes this function but you specifically commented `// do not follow symlinks`... should i catch the symlink and do something different logic?

deitch:
Why did you pick the actions commit in, as opposed to just rebasing?

deitch:
I am not sure I understand the question, but the symlink is because we are building information about the file itself, not the file it points to, so we need _not_ to follow symlinks. Think of it like this: if I wanted to make an exact copy of the contents of one directory into another (which is what we are doing here, except instead of copying from one to another, we are building a new filesystem with the same contents), it wouldn't be a true copy if the source had 3 files and a symlink, while the destination had 4 files: the original 3, and the target of the symlink. It would need to be 3 files and a symlink.

Does `times.Lstat` handle it differently than `os.Lstat`?

deitch:
My tests show that it does the right thing. Look at [this gist](https://gist.github.com/deitch/f6bef29d3c121b461691009bbfdee8f1).

luthermonson:
sorry for not being clear, im just trying to not break anything as to be perfectly honest i know nothing about any of these formats and I don't want to accidentally make anything worse and will go with whatever you tell me is the right way to do this.

RE The extra commit, i wasn't paying attention and will fix.

RE LStat: The times and os Lstat did what I expected and tests pass, I don't know if it's breaking anything further by doing it that way. I will fix my PR and remove the extra commit and keep the LStas, if you think this needs more just tell me and I'll do it.

luthermonson:
@deitch oh i figured it out, i changed the os.Lstat you had to an os.Stat on accident while adding the times.Stat which needed to be a times.Lstat. So, I broke it and didn't realize it. I'm closing this PR and I have a new branched from master with Lstat's coming shortly. I'm sorry but between starting this and finishing this I've been on vacation 1.5wks and sort of forgot what I did and made this mistake. New PR coming.

PR #29, opened 2019-10-25
Adding Times Package

luthermonson:
Moving logic to get atime and ctime to a package with more architecture support.


luthermonson:
@deitch much cleaner and Lstat works well. Forgive me for the confusion.

deitch:
Yes, quite nice. Thanks!

PR #31, opened 2019-10-28
Split Stat_t for windows

luthermonson:
Not sure this is the actual fix but my windows compiles are solid now with this change locally. If you want changes ask and I'll fix.

### DIFF
--- a/filesystem/iso9660/rockridge.go
+++ b/filesystem/iso9660/rockridge.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"os"
 	"sort"
-	"syscall"
 	"time"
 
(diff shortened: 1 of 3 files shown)


deitch:
I see no issues with this (and thank you, indeed).

Is the issue that Windows was returning `sys != nil` and so was trying to do it, when it should just be `uint32(0)` on Windows?

luthermonson:
@deitch issue was syscall.Stat_t doesn't exist in windows (https://golang.org/src/syscall/zsysnum_windows_amd64.go) they just ignored it so not much else we can do except return 0s. I have no idea what that will do to ISOs built on windows tbh but since you defaulted it to 0s I assume it doesn't hurt anyones feelings.

PR #39, opened 2019-11-13
Add iso example

### LINKED ISSUE #37: How to create a simple iso image
Hi,
I tried to follow the documentation on godoc. But I was not successful to create a simple iso image.
I tried this: https://gist.github.com/kstieger/3fa5b945a9008de91c974ddd611ab325#file-myiso-go

The iso file is generated, but when I try to mount it, I receive a error message:
```
$ mount -o loop /tmp/mydisk.iso /media/
mount: /media: wrong fs type, bad option, bad superblock on /dev/loop0, missing codepage or helper program, or other error.
```

Thanks a lot,
Kurt

deitch:
Fixes #37 

Do you mind that I based the example on your code @kstieger? Please comment here.

### DIFF
--- a/examples/iso_create.go
+++ b/examples/iso_create.go
@@ -1,0 +1,44 @@
+package main
+
+import (
+	"fmt"
+	diskfs "github.com/diskfs/go-diskfs"
+	"github.com/diskfs/go-diskfs/disk"
+	"github.com/diskfs/go-diskfs/filesystem"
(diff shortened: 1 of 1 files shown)


deitch:
I may have modified it a bit here and there, but it was a good base and forced me to put in place a good example, while fixing a bug. Thank you.

PR #43, opened 2020-01-09
introduce OpenWithMode()

guimagalhaes:
Here I propose a new function to open a disk device with a user selected mode: the OpenWithMode() function.

With this function, clients are able to execute read-only operations in situations where the device is already open for write. For instance, LVM physical volumes will already be open for write and then this library does not support this case which is a very common case these days.
Clients can also make use of the library in a safer mode when just reading the current disk partitioning format is required and then make sure no change to the disk will be allowed.

The proposed changes adds a new flag to the Disk structure to indicate it is writable or not. This flag is used to explicitly fail any operation which would modify the disk if it was opened in read-only mode.

The original interface is not broken and then existing clients are not impacted.

* introduce OpenWithMode()

Signed-off-by: Guilherme Magalhaes <guilherme.magalhaes@hpe.com>

* add tests with readonly disks for write operations

Signed-off-by: Guilherme Magalhaes <guilherme.magalhaes@hpe.com>

### DIFF
--- a/disk/disk.go
+++ b/disk/disk.go
@@ -25,6 +25,7 @@ type Disk struct {
 	LogicalBlocksize  int64
 	PhysicalBlocksize int64
 	Table             partition.Table
+	Writable          bool
 }
 
 // Type represents the type of disk this is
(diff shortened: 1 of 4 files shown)


deitch:
I like this. Thanks for doing it, and for adding tests.

PR #55, opened 2020-03-04
Update fat32.go

ActNotSign:
fixed, invalid cluster chain：can't read directory


ActNotSign:
> Why are we doing this? This function should not print; if there is an error, it should return it. What problem are you trying to solve?

If return error, we can't read directyory list. So , i think the problem is a warning.

deitch:
The first cluster in a fat32 is always supposed to be 2. If the chain is <2, the fat is messed up and shouldn't be readable.

Let's turn it the other way. Do you have a replicable issue?

ActNotSign:
> The first cluster in a fat32 is always supposed to be 2. If the chain is <2, the fat is messed up and shouldn't be readable.
> 
> Let's turn it the other way. Do you have a replicable issue?

Issue #56, opened 2020-03-12
lsblk does not show partitions after writing GPT partitions from docker container


akhilerm:
**What happened**
Created a single partition on a 14GB disk using the below code snippet. The issue occurred when the partition was created from a docker container. When executed directly it worked as expected.
```
table := &gpt.Table{
		LogicalSectorSize: 512,
		ProtectiveMBR:     true,
	}
partition := &gpt.Partition{
		Start: 2048,
		End:   30277598,
		Type:  gpt.LinuxFilesystem,
	}
table.Partitions = append(table.Partitions, partition)
p := "/dev/sdb"
d,err := diskfs.Open(p)
err = d.Partition(table)
```
`lsblk` does not show the newly created partition `/dev/sdb1`.
```
sda          8:0    0 465.8G  0 disk 
├─sda1       8:1    0   487M  0 part /boot/efi
├─sda2       8:2    0     1K  0 part 
├─sda5       8:5    0 114.5G  0 part /
├─sda6       8:6    0 339.4G  0 part /home
└─sda7       8:7    0  11.5G  0 part [SWAP]
sdb          8:16   1  14.4G  0 disk 
``` 
But `fdisk -l /dev/sdb` shows that  a partition is present.
```
Disk /dev/sdb: 14.4 GiB, 15502147584 bytes, 30277632 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: gpt
Disk identifier: 4B33D1CC-0804-4A47-B237-D19CEA71F7E1

Device     Start      End  Sectors  Size Type
/dev/sdb1   2048 30277598 30275551 14.4G Linux filesystem
```

**What I think the problem is?**
Partition utilities like `fdisk` forces the kernel to reread the partition table after partition blocks are written, so the partition table gets updated immediately. If the table is not re-read, the time at which kernel probes the disk and updates cannot be determined.

**Proposed solution**
diskfs should use a `ioctl(fd, BLKRRPART)` call to force the kernel to re-read the partition table after the partition table is written onto the disk.
This needs to be done only in the case of blockdevices, while working with files this need not be the case.
Cons of the solution: It will make `go-diskfs` platform dependent as `ioctl()` is unix specific.

PR #57, opened 2020-03-13
force kernel to reread partition table

akhilerm:
force kernel to reread partition table after successfully writing a partition table to the disk. This is done by an ioctl call with BLKRRPART request

Fixes : #56 

Signed-off-by: Akhil Mohan <akhil.mohan@mayadata.io>

### DIFF
--- a/disk/disk.go
+++ b/disk/disk.go
@@ -7,6 +7,7 @@ package disk
 import (
 	"errors"
 	"fmt"
+	"golang.org/x/sys/unix"
 	"io"
 	"os"
 
@@ -36,6 +37,8 @@ const (
 	File Type = iota
 	// Device is an OS-managed block device
 	Device
+	// blkRRPart is the ioctl request to re-read the partition table
+	blkRRPart = 0x0000125F
 )
 
 var (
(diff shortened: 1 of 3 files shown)


disk/disk.go
```diff
@@ -204,3 +213,16 @@ func (d *Disk) GetFilesystem(partition int) (filesystem.FileSystem, error) {
...
 	return nil, fmt.Errorf("Unknown filesystem on partition %d", partition)
 }
+
+// ReReadPartitionTable forces the kernel to re-read the partition table
+// on the disk.
+//
+// It is done via an ioctl call with request as BLKRRPART.
+func (d *Disk) ReReadPartitionTable() error {
+	fd := d.File.Fd()
+	_, err := unix.IoctlGetInt(int(fd), blkRRPart)
```
deitch: Please replace the local const `blkRRPart` with the official `unix.BLKRRPART`.

akhilerm:
Is it necessary that we use separate files for linux and darwin. We could use a single `disk_unix.go` with the build tags,
```
// +build aix darwin dragonfly freebsd linux netbsd openbsd solaris
```
and then a separate `disk_windows.go` that just returns nil.

deitch:
Sure. Sounds good.

akhilerm:
Done. made the changes.

disk/disk.go
```diff
@@ -36,6 +36,8 @@ const (
 	File Type = iota
 	// Device is an OS-managed block device
 	Device
+	// blkRRPart is the ioctl request to re-read the partition table
+	blkRRPart = 0x0000125F
```
deitch: Should get rid of this.

deitch:
One piece left to remove. Any idea why tests are still failing?

disk/disk.go
akhilerm: Done.

akhilerm:
Done. The tests has passed.

deitch:
Thanks!

PR #64, opened 2020-08-19
Fix root folder clusters

rtreffer:
While trying to generate an RPI boot image I noticed that the library fails to create a root directory with all files.

After adding a failing test case (woohoo) I discovered that the `allocateSpace` function needs to be called with a file size and the beginning of the current chain (otherwise the size delta is not computable).
Also wiring up the chain was problematic.

After applying these changes the 32 file root directory test passes.

EDIT: there is also a drive-by commit: minClusterSize and maxClusterSize were swapped :see_no_evil:

### DIFF
--- a/filesystem/fat32/fat32.go
+++ b/filesystem/fat32/fat32.go
@@ -41,8 +41,8 @@ type SectorSize uint16
 const (
 	// SectorSize512 is a sector size of 512 bytes, used as the logical size for all FAT filesystems
 	SectorSize512        SectorSize = 512
-	maxClusterSize       int        = 128
-	minClusterSize       int        = 65529
+	minClusterSize       int        = 128
+	maxClusterSize       int        = 65529
 	bytesPerSlot         int        = 32
 	maxCharsLongFilename int        = 13
 )
--- a/filesystem/fat32/fat32.go
+++ b/filesystem/fat32/fat32.go
@@ -584,7 +584,7 @@ func (fs *FileSystem) getClusterList(firstCluster uint32) ([]uint32, error) {
 		switch {
 		case fs.table.isEoc(newCluster):
 			complete = true
-		case cluster <= 2:
+		case cluster < 2:
 			return nil, fmt.Errorf("Invalid cluster chain at %d", cluster)
 		}
 		cluster = newCluster
@@ -641,9 +641,9 @@ func (fs *FileSystem) writeDirectoryEntries(dir *Directory) error {
 	if err != nil {
 		return fmt.Errorf("Unable to get clusters for directory: %v", err)
 	}
-	extraClusters := len(b)/(int(fs.bootSector.biosParameterBlock.dos331BPB.dos20BPB.sectorsPerCluster)*fs.bytesPerCluster) - len(clusterList)
-	if extraClusters > 0 {
-		clusters, err := fs.allocateSpace(uint64(extraClusters), clusterList[len(clusterList)-1])
+
+	if len(b) > len(clusterList)*fs.bytesPerCluster {
+		clusters, err := fs.allocateSpace(uint64(len(b)), clusterList[0])
 		if err != nil {
 			return fmt.Errorf("Unable to allocate space for directory entries: %v", err)
 		}
--- a/filesystem/fat32/fat32_test.go
+++ b/filesystem/fat32/fat32_test.go
@@ -516,6 +516,72 @@ func TestFat32OpenFile(t *testing.T) {
 		})
 	})
 
+	// write many files to exceed the first cluster, then read back
+	t.Run("Write Many", func(t *testing.T) {
+		runTest := func(t *testing.T, pre, post int64) {
+			f, err := tmpFat32(false, pre, post)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if keepTmpFiles == "" {
+				defer os.Remove(f.Name())
+			} else {
+				fmt.Println(f.Name())
+			}
+			fileInfo, err := f.Stat()
+			if err != nil {
+				t.Fatalf("Error getting file info for tmpfile %s: %v", f.Name(), err)
+			}
+			fs, err := fat32.Create(f, fileInfo.Size()-pre-post, pre, 512, " NO NAME")
+			if err != nil {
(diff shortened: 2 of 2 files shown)


filesystem/fat32/fat32.go
```diff
@@ -41,8 +41,8 @@ type SectorSize uint16
 const (
 	// SectorSize512 is a sector size of 512 bytes, used as the logical size for all FAT filesystems
 	SectorSize512        SectorSize = 512
-	maxClusterSize       int        = 128
-	minClusterSize       int        = 65529
+	minClusterSize       int        = 128
+	maxClusterSize       int        = 65529
```
deitch, line 45: Oh Lordy, this one is painful! I cannot believe this lived all this time.

filesystem/fat32/fat32.go
```diff
@@ -631,7 +631,7 @@ func (fs *FileSystem) writeDirectoryEntries(dir *Directory) error {
 	}
 	extraClusters := len(b)/(int(fs.bootSector.biosParameterBlock.dos331BPB.dos20BPB.sectorsPerCluster)*fs.bytesPerCluster) - len(clusterList)
 	if extraClusters > 0 {
-		clusters, err := fs.allocateSpace(uint64(extraClusters), clusterList[len(clusterList)-1])
+		clusters, err := fs.allocateSpace(uint64(len(b)), clusterList[0])
```
deitch, line 634: I assume this was what you mean when you wrote:

> function needs to be called with a file size and the beginning of the current chain (otherwise the size delta is not computable)

Can you explain these?

* Clearly, passing the `extraClusters` was wrong, since it should pass the bytes required. But this would pass it the entire `b`, when all we need is the excess. 
* Why do we pass it the previous as the first, i.e. `clusterList[0]`, and not the last?

filesystem/fat32/fat32.go
```diff
@@ -810,21 +810,21 @@ func (fs *FileSystem) allocateSpace(size uint64, previous uint32) ([]uint32, err
...
+			continue
+		}
+		cur := allocated[i-1]
+		fs.table.clusters[cur] = next
 	}
-	for i := 0; i < lastAlloc; i++ {
-		allClusters[allocated[i]] = allocated[i+1]
+	fs.table.clusters[allocated[len(allocated)-1]] = fs.table.eocMarker
+	if len(clusters) > 0 {
+		fs.table.clusters[clusters[len(clusters)-1]] = allocated[0]
```
deitch, line 826: It looks like you are rejiggering the tables here? Why not just append it to the end?

deitch (commented):
Thanks for submitting this, and catching some pretty glaring errors. I don't fully get some of the changes.

filesystem/fat32/fat32.go
rtreffer: It is possible to call with len and let it load the chain, or call with the delta and the last sector. I'll check if I can simplify this, I think I might have overshoot the goal a bit trying to fix this

rtreffer:
@deitch you are right, I've removed the changes to allocateSpace. There is just one catch that the size is the total size including the last sector, so a high chance for off-by-one. But it works, the diff is now pretty small (except for the testcase)

rtreffer:
Ok, `allocateSpace` has a different bug: if previous points to the first cluster then the current implementation will link the first block with the new allocation chain. That's why it is needed for #65

I've extracted that into https://github.com/diskfs/go-diskfs/pull/69 - my rework fixed that issue. I will rework #65 on top of that fix.

deitch:
At this point, there is just that line swap, and a bunch of tests. I am more than happy to merge in the line swap, but are you waiting to do more here?

filesystem/fat32/fat32.go
```diff
@@ -572,7 +572,7 @@ func (fs *FileSystem) getClusterList(firstCluster uint32) ([]uint32, error) {
 		switch {
 		case fs.table.isEoc(newCluster):
 			complete = true
-		case cluster <= 2:
+		case cluster < 2:
```
rtreffer, line 575: @deitch this change is critical because block 2 can be a chain

rtreffer:
The cluster 2 is currently not allowed to be a cluster, that's critical to make multi-cluster root folder work.

filesystem/fat32/fat32.go
deitch: Makes sense.

filesystem/fat32/fat32.go
```diff
@@ -631,7 +631,8 @@ func (fs *FileSystem) writeDirectoryEntries(dir *Directory) error {
 	}
 	extraClusters := len(b)/(int(fs.bootSector.biosParameterBlock.dos331BPB.dos20BPB.sectorsPerCluster)*fs.bytesPerCluster) - len(clusterList)
 	if extraClusters > 0 {
-		clusters, err := fs.allocateSpace(uint64(extraClusters), clusterList[len(clusterList)-1])
+		extraBytes := extraClusters * fs.bytesPerCluster
+		clusters, err := fs.allocateSpace(uint64(fs.bytesPerCluster + extraBytes), clusterList[len(clusterList)-1])
```
deitch, line 635: Agreed that the previous `allocateSpace(extraClusters` was wrong, since the first arg to `allocateSpace` is the file size. But then the above is incorrect too. You are just passing it the `bytesPerCluster` and the `extraBytes`. Shouldn't we just be passing it `len(b)`?

rtreffer: if we pass len(b) then we would also have to pass the first cluster and extraClusters is not needed at all. I can switch to that again.

deitch: Here is the comment on `allocateSpace()`:

```go
// allocateSpace ensure that a cluster chain exists to handle a file of a given size.
// arguments are file size in bytes and starting cluster of the chain
// if starting is 0, then we are not (re)sizing an existing chain but creating a new one
// returns the indexes of clusters to be used in order
```

The file is of a given size, and starting cluster of the chain. So we should stick to that.

rtreffer: Fixed, needed a rebase to use the fixed `allocateSpace` method.

deitch (approved):

PR #65, opened 2020-08-20
Fix streaming fat32 file write

rtreffer:
This PR (based on #64 ) fixes streaming fat32 file writes.

After each write the file offset should be advanced by the number of bytes written. see commit dd19876

### DIFF
--- a/filesystem/fat32/file.go
+++ b/filesystem/fat32/file.go
@@ -19,7 +19,8 @@ type File struct {
 // It returns the number of bytes read and any error encountered.
 // At end of file, Read returns 0, io.EOF
 // reads from the last known offset in the file from last read or write
-// use Seek() to set at a particular point
+// and increments the offset by the number of bytes read.
+// Use Seek() to set at a particular point
 func (fl *File) Read(b []byte) (int, error) {
 	// we have the DirectoryEntry, so we can get the starting cluster location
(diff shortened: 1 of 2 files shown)


deitch:
Will wait until we finish #64 to comment on this.

rtreffer:
@deitch I've reworked this PR on top of #69 which should be easier to review

deitch:
#69 is merged in, so this should be rebased. It will reduce the number of changes.

filesystem/fat32/file.go
```diff
@@ -154,6 +156,9 @@ func (fl *File) Write(p []byte) (int, error) {
 		file.WriteAt(p[totalWritten:totalWritten+toWrite], int64(offset)+fs.start)
 		totalWritten += toWrite
 	}
+
+	fl.offset = fl.offset + int64(totalWritten)
```
deitch, line 160: Personally, I think that Unix should have separate offsets for read and write. But we are not about to change 40+ years of history here, and we should stick to the standard, so this makes sense.

deitch (changes requested):
Just needs a rebase and this can go in

rtreffer:
freshly rebased

deitch:
Nothing like some fresh rebase with a mug of tea for breakfast. 😀

deitch (approved):

PR #66, opened 2020-08-20
Fix GPTProtective partition type (0xee) for hybrid/protective MBRs

rtreffer:
From https://en.wikipedia.org/wiki/Partition_type#PID_EEh
```
0xEE GPT protective MBR
0xEF EFI system partition
```


deitch (approved):
Nice catch. It didn't make sense to have 2 `0xef` in a row anyways. Slip of the finger on the keyboard, I guess.

PR #68, opened 2020-08-24
Correctly write backup blocks

rtreffer:
Fixes fsck.vfat complais about the backup blocks:
- boot sector is copied to the backup block location
- filesystem information is packed into the next logical block
- fat backup is written alongside the primary copy

fsck.vfat does not complain about these issues after applying the patch

### DIFF
--- a/filesystem/fat32/fat32.go
+++ b/filesystem/fat32/fat32.go
@@ -228,6 +228,17 @@ func Create(f util.File, size int64, start int64, blocksize int64, volumeLabel s
 		return nil, fmt.Errorf("Wrote %d bytes of MS-DOS Boot Sector to disk instead of expected %d", count, SectorSize512)
 	}
 
+	// write backup to the file
+	if fsisBackupSector > 0 {
+		count, err = f.WriteAt(b, int64(fsisBackupSector)*int64(SectorSize512)+int64(start))
+		if err != nil {
+			return nil, fmt.Errorf("Error writing MS-DOS Boot Sector to disk: %v", err)
+		}
+		if count != int(SectorSize512) {
+			return nil, fmt.Errorf("Wrote %d bytes of MS-DOS Boot Sector to disk instead of expected %d", count, SectorSize512)
+		}
+	}
+
 	// boot sector is in place
 
 	// create and allocate FAT32 FSInformationSector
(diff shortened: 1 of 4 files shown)


filesystem/fat32/fat32.go
```diff
@@ -228,6 +228,17 @@ func Create(f util.File, size int64, start int64, blocksize int64, volumeLabel s
 		return nil, fmt.Errorf("Wrote %d bytes of MS-DOS Boot Sector to disk instead of expected %d", count, SectorSize512)
 	}
 
+	// write backup to the file
+	if fsisBackupSector > 0 {
```
deitch, line 232: I don't get this. Why are we writing the boot sector bytes to the FSIS Backup Sector?

deitch: And then we write it again later?

filesystem/fat32/fat32.go
```diff
@@ -241,10 +252,11 @@ func Create(f util.File, size int64, start int64, blocksize int64, volumeLabel s
 		return nil, fmt.Errorf("Could not create a valid byte stream for a FAT32 Filesystem Information Sector: %v", err)
 	}
 	fsisPrimary := int64(fsisPrimarySector * uint16(SectorSize512))
-	fsisBackup := int64(fsisBackupSector * uint16(SectorSize512))
 
 	f.WriteAt(fsisBytes, fsisPrimary+int64(start))
-	f.WriteAt(fsisBytes, fsisBackup+int64(start))
+	if fsisBackupSector > 0 {
+		f.WriteAt(fsisBytes, int64(fsisBackupSector+1)*int64(SectorSize512)+int64(start))
```
deitch, line 258: Here we write it again, with the FSIS bytes, but we write it one sector _after_ the FSIS backup sector?

deitch (commented):
Some things I didn't fully get.

filesystem/fat32/fat32.go
rtreffer: It looks like this is actually the backup boot sector
> For FAT32 file systems, the reserved sectors include a File System Information Sector at logical sector 1 and a Backup Boot Sector at logical sector 6.
see e.g. https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system

I have to double check if the FSIS is added there, but I have seen references that 3 backup blocks are written, which would be the FSIS on change (here) and above

rtreffer: Ah, later on the same wikipedia page
> First logical sector number of a copy of the three FAT32 boot sectors, typically 6.

So it's not a backup of the FSIS but of the whole boot sector

rtreffer: Should the struct member be renamed, too?

rtreffer:
@deitch I went ahead and renamed backupFSInfoSector to backupBootSector because that's what it is. I hope that clarifies the issue.

deitch:
Everything you write here makes sense based on that page. I have been trying hard to recall where we got this, but I guess it doesn't matter as long as we get it right.

deitch (approved):

PR #69, opened 2020-08-24
Fix allocateSpace if previous points to the beginning of a chain

rtreffer:
Allocate space will link the old chain with the new chain. This can only
work if previous is pointing to the last cluster of the privious chain.


deitch (approved):

deitch:
You are having some serious fun with fat32. I am curious what you are using it for.

PR #70, opened 2020-08-25
Fix cluster location of .. directory entry

rtreffer:
The cluster location of parent directory entries (`..`) pointing to the root directory (`/`) must be 0, not 2. `fsck.vfat` complains about this.

Currently based on #64 for testing purposes (large root folder)

### DIFF
--- a/filesystem/fat32/fat32.go
+++ b/filesystem/fat32/fat32.go
@@ -734,11 +734,16 @@ func (fs *FileSystem) readDirWithMkdir(p string, doMake bool) (*Directory, []*di
 					return nil, nil, fmt.Errorf("Failed to create subdirectory %s", "/"+strings.Join(paths[0:i+1], "/"))
 				}
 				// make a basic entry for the new subdir
+				parentDirectoryCluster := currentDir.clusterLocation
+				if parentDirectoryCluster == 2 {
+					// references to the root directory (cluster 2) must be stored as 0
+					parentDirectoryCluster = 0
+				}
 				dir := &Directory{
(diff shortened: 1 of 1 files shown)


deitch:
This needs a clean rebase, now that #64 is in.

deitch:
> must be 0, not 2

This doesn't make sense to me. Why would it point to the 0, if it should be pointing at 2? I can see it breaking all sorts of things. Is this in the spec somewhere?

Nonetheless, I did just create a filesystem image using standard tools and read the entries, and they respect exactly what you said.

Does it break anything else? I am concerned that there might be other areas in the code that might try to look it up and error out? I cannot think of anything that reads the parent off-hand.

rtreffer:
It makes no sense, I found it at https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system
> Entries with the Volume Label flag, subdirectory ".." pointing to the FAT12 and FAT16 root, and empty files with size 0 should have first cluster 0. 

I think it is still a leftover from that. Given that the cluster 0 is the boot sector + fsis which points to the root directory.

So half-way reasonable, but a horribly exception. `fsck.vfat` was fixing this so I assume it is safe to use.

deitch:
> So half-way reasonable, but a horribly exception. fsck.vfat was fixing this so I assume it is safe to use.

Same here. Once rebased, we can get it in

rtreffer:
Freshly rebased

deitch (approved):
Thank for finding and fixing this

PR #71, opened 2020-08-25
Fix duplicated 8+3 filenames for long file name entries

rtreffer:
We need two pieces to generate unique 8+3 names from arbitrary long file names: a fast lookup map of existing names to find collisions and a way to generate names in the case of a collision.

The lookup map is lazy generated on first entry addition. This means the normal loading logic does not degrade in performance.

Several existing truncation approaches are listed on https://en.wikipedia.org/wiki/8.3_filename#VFAT_and_Computer-generated_8.3_filenames:
1. Shorten the name to 6 characters, then append `~1` .. `~9`
1. Shorten the name to 5 characters, then append `~10`..`~99`
1. Shorten the name to 2 characters, append 4 characters from the hash of the original filename

An example of a problematic folder would be the Big Buck Bunny PNG download (see https://media.xiph.org/BBB/BBB-360-png/). All files have the same prefix and numbering those has quadratic cost.
The hashing is usually used as a runtime complexity limiter.

This PR implements a variation of `1.+3.+random`, where random is a simple random number based file name generation.

**WIP**: I am still not truly happy with the collision avoidance method, I would like to get some feedback on better / easier avoidance method.

After applying this patch and the other open PRs `fsck.vfat` finds only one minor issue on a complex filled disk: the go-diskfs fat32 implementation does not track the number of free clusters. Everything else looks good from a on-disk layout perspective.

### DIFF
--- a/filesystem/fat32/directoryentry.go
+++ b/filesystem/fat32/directoryentry.go
@@ -3,7 +3,10 @@ package fat32
 import (
 	"encoding/binary"
 	"fmt"
+	"hash/fnv"
+	"math/rand"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 )
@@ -382,7 +385,7 @@ func calculateSlots(s string) int {
 // returns shortName, extension, isLFN, isTruncated
 //   isLFN : was there an LFN that had to be converted
 //   isTruncated : was the shortname longer than 8 chars and had to be converted?
-func convertLfnSfn(name string) (string, string, bool, bool) {
+func convertLfnSfn(name string, cache map[string]bool) (string, string, bool, bool) {
 	isLFN, isTruncated := false, false
 	// get last period in name
 	lastDot := strings.LastIndex(name, ".")
@@ -413,11 +416,47 @@ func convertLfnSfn(name string) (string, string, bool, bool) {
 		isLFN = true
 	}
 
+	truncate := func(name string, n int) string {
+		return name[0:6] + "~" + strconv.Itoa(n)
+	}
+
+	hash := func(shortName string, n int) string {
+		h := fnv.New32a()
+		// the input name may have more information than the short name
+		h.Write([]byte(name))
+		return shortName[0:2] + fmt.Sprintf("%04X", (h.Sum32()%0x10000)) + "~" + strconv.Itoa(n)
+	}
+
+	random := func() string {
+		return fmt.Sprintf("%06X", (rand.Int63()%0x1000000)) + "~" + strconv.Itoa(1+rand.Intn(8))
+	}
+
 	// convert shortName to 8 chars
 	if len(shortName) > 8 {
 		isLFN = true
 		isTruncated = true
-		shortName = shortName[:6] + "~" + "1"
+		if !cache[truncate(shortName, 9)+"."+extension] {
+			i := 1
+			for i < 9 && cache[truncate(shortName, i)+"."+extension] {
+				i++
+			}
+			shortName = truncate((shortName), i)
+		} else {
+			i := 0
+			for i <= 9 && cache[hash(shortName, i)+"."+extension] {
+				i++
+			}
+			if i <= 9 {
+				shortName = hash(shortName, i)
+			} else {
+				rnd := random()
+				for i < 1000 && cache[rnd+"."+extension] {
+					i++
+					rnd = random()
+				}
+				shortName = rnd
+			}
+		}
 	}
 	return shortName, extension, isLFN, isTruncated
 }
--- a/filesystem/fat32/directory.go
+++ b/filesystem/fat32/directory.go
@@ -7,7 +7,8 @@ import (
 // Directory represents a single directory in a FAT32 filesystem
 type Directory struct {
 	directoryEntry
-	entries []*directoryEntry
+	sfnCache map[string]bool
+	entries  []*directoryEntry
 }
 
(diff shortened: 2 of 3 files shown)


deitch:
To be clear, you basically are addressing [this todo](https://github.com/diskfs/go-diskfs/blob/b1a4e065bf7ac4ecbe331c1b182d2d5b4c380805/filesystem/fat32/directory.go#L44) ?

rtreffer:
Exactly. The issue at hand is that if I write a raspberry pi boot folder then there are many dtbs with similar prefixes (bmc27...-). This causes those to be written out with the same short name and `fsck.vfat` regards the same short name as the same file, thus complaining about "duplicated directory entry".

The wiki page has a few nice pointers on how to solve this. The `~1` .. `~9` is very human friendly IMHO. The hashing is less friendly but avoids runtime complexity issues. Both will fall apart at >10k directory entries as they at some point run out of possible names, thus random as a last resort....

The basic hashmap of all entries should be less controversal and easy to add, I might extract the collision avoiding SFN search from the rest of the method body, that should make it well contained.

deitch:
> This PR implements a variation of 1.+3.+random, where random is a simple random number based file name generation.

I didn't understand what you wrote here. It looks (from the code) like you do the following:

1. Try to to the first option, i.e. "first 6 chars + tilde + sequential digit", eventually using up 9 slots. This is what the Windows algorithm does. If all of those 9 slots are taken
1. Try to do the Windows NT (and later) hashing
1. Do something with random

> WIP: I am still not truly happy with the collision avoidance method, I would like to get some feedback on better / easier avoidance method.

To what do you refer?

rtreffer:
I am referring to the fact that
1. I would like to know what you think :-)
1. I am not happy with the implementation yet
The hashing of NT and onwards is unknown, and the wikipedia page does not list an avoidance algorithm, but there must be one as collisions starts to become likely for a 16 bit number if you go beyond a few hundred files.

deitch:
> The hashing of NT and onwards is unknown, and the wikipedia page does not list an avoidance algorithm, but there must be one as collisions starts to become likely for a 16 bit number if you go beyond a few hundred files

What I think? Pretty simple, actually.

* What we have now doesn't handle even a single conflict. Which means _anything_ is better. Even just the 9-file-limitation of `[0-9A-Z]~[1-9]` would be a material improvement.
* Adding the second stage of hashing is even better, as it now gets us somewhere in the range of hundreds or possibly thousands of files before a likely collision

In all seriousness, this would take us from a guaranteed collision with >1 file, to a guaranteed collision with >9 file (an order of magnitude improvement) to a possible collision with >hundreds of files (a 2+ order of magnitude improvement). It isn't perfect? So what.

Now that I understand it, I like most everything in here. I have some suggestions on the coding style, but will put those inline.

filesystem/fat32/directoryentry.go
```diff
@@ -413,11 +416,47 @@ func convertLfnSfn(name string) (string, string, bool, bool) {
 		isLFN = true
 	}
 
+	truncate := func(name string, n int) string {
+		return name[0:6] + "~" + strconv.Itoa(n)
+	}
```
deitch, line 421: I am not madly in love with inline anonymous functions here, as they will get declared every time we call `convertLfnSfn`, and they are harder to understand and test.

Can we pull these three out into named functions, and give them more representative names?

filesystem/fat32/directoryentry.go
```diff
@@ -413,11 +416,47 @@ func convertLfnSfn(name string) (string, string, bool, bool) {
...
+	hash := func(shortName string, n int) string {
+		h := fnv.New32a()
+		// the input name may have more information than the short name
+		h.Write([]byte(name))
+		return shortName[0:2] + fmt.Sprintf("%04X", (h.Sum32()%0x10000)) + "~" + strconv.Itoa(n)
+	}
+
+	random := func() string {
+		return fmt.Sprintf("%06X", (rand.Int63()%0x1000000)) + "~" + strconv.Itoa(1+rand.Intn(8))
+	}
```
deitch, line 432: `random` is a bit too generic of a name, and probably will confuse. We should pull this out too, but make it more specifically-named.

filesystem/fat32/directoryentry.go
```diff
@@ -413,11 +416,47 @@ func convertLfnSfn(name string) (string, string, bool, bool) {
...
 	if len(shortName) > 8 {
 		isLFN = true
 		isTruncated = true
-		shortName = shortName[:6] + "~" + "1"
+		if !cache[truncate(shortName, 9)+"."+extension] {
+			i := 1
+			for i < 9 && cache[truncate(shortName, i)+"."+extension] {
+				i++
+			}
+			shortName = truncate((shortName), i)
```
deitch, line 443: This calls `truncate()` duplicatively several times. Can we tighten it up a bit?

Also, why is this:

```go
			i := 1
 			for i < 9 && cache[truncate(shortName, i)+"."+extension] {
 				i++
 			}
```

and not the simpler and more idiomatic:

```go
			var found string
 			for i := i; i++; i < 9 {
                            found = truncate(shortName, i)+"."+extension
                            if _, ok := cache[found]; ok {
                               break
                            }
 			}
                        shortName = found
```

or similar

deitch (changes requested):
Just some stylistic questions, but this really is quite good.

PR #81, opened 2020-12-11
Finish reading the entire file in iso9660

carbonin:
`size` here is the amount we have left to write. To determine if we
are done, we need to compare the new offset to the total size of
the file.

Without this fix we would always bail at around halfway
(+/- the buffer size) through any file larger than the buffer size.

### DIFF
--- a/filesystem/iso9660/common_internal_test.go
+++ b/filesystem/iso9660/common_internal_test.go
@@ -46,3 +46,32 @@ func GetTestFile(t *testing.T) (*File, string) {
 		offset:         0,
 	}, "README\n"
 }
+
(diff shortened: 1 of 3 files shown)


carbonin:
Without the fix, the new test fails like this:
```
--- FAIL: TestLargeFileCopy (0.00s)
    file_test.go:38: copied 2629632 bytes instead of expected 5242880
FAIL
exit status 1
FAIL	github.com/diskfs/go-diskfs/filesystem/iso9660	0.826s
```

carbonin:
cc @deitch

deitch (approved):

PR #88, opened 2021-01-03
squashfs

deitch:


deitch:
This actually is passing. Something strange is happening with the tests, related to docker run. Can run all cleanly locally.

PR #91, opened 2021-01-22
Remove the iso workspace after running Finalize

### LINKED ISSUE #90: iso finalize leaks the workspace temp directory
Currently this library leaks temp directories when used as shown in the [iso_create.go example](https://github.com/diskfs/go-diskfs/blob/1236d9d5ae8d2ecdc8cd0eeee3cf7dc9d489dca7/examples/iso_create.go)

The `Finalize` method also removes the reference to the workspace, so what is the best practice here?

Should the example be changed to something like this:
```patch
diff --git a/examples/iso_create.go b/examples/iso_create.go
index c0983f8..73c6f91 100644
--- a/examples/iso_create.go
+++ b/examples/iso_create.go
@@ -33,6 +33,9 @@ func CreateIso(diskImg string) {
        if !ok {
                check(fmt.Errorf("not an iso9660 filesystem"))
        }
+       ws := fs.Workspace()
        err = iso.Finalize(iso9660.FinalizeOptions{})
        check(err)
+       err = os.RemoveAll(ws)
+       check(err)
 }
```

Or should finalize remove the workspace before it clears the reference in the filesystem struct?

I'll make whichever change you think makes more sense @deitch

carbonin:
Previously this directory which contained all the files in the iso
would be left in the temp directory.

Fixes #90

### DIFF
--- a/filesystem/iso9660/finalize.go
+++ b/filesystem/iso9660/finalize.go
@@ -713,6 +713,8 @@ func (fs *FileSystem) Finalize(options FinalizeOptions) error {
 	b = terminator.toBytes()
 	f.WriteAt(b, int64(location)*int64(blocksize))
 
+	_ = os.RemoveAll(fs.workspace)
+
 	// finish by setting as finalized
 	fs.workspace = ""
 	return nil
--- a/filesystem/iso9660/finalize_test.go
+++ b/filesystem/iso9660/finalize_test.go
@@ -328,10 +328,17 @@ func TestFinalizeRockRidge(t *testing.T) {
 			}
 		}
 
+		workspace := fs.Workspace()
+
 		err = fs.Finalize(iso9660.FinalizeOptions{RockRidge: true})
 		if err != nil {
 			t.Fatal("Unexpected error fs.Finalize({RockRidge: true})", err)
 		}
+
+		if _, err := os.Stat(workspace); !os.IsNotExist(err) {
+			t.Fatalf("Workspace dir %s should be removed after finalizing", workspace)
+		}
+
 		// now need to check contents
 		fi, err := f.Stat()
 		if err != nil {


filesystem/iso9660/finalize.go
```diff
@@ -713,6 +713,8 @@ func (fs *FileSystem) Finalize(options FinalizeOptions) error {
 	b = terminator.toBytes()
 	f.WriteAt(b, int64(location)*int64(blocksize))
 
+	os.RemoveAll(fs.workspace)
```
carbonin, line 716: Should we report an error here even if the iso was created successfully?

carbonin:
@deitch can you take a look when you get a chance?

cc @danielerez

filesystem/iso9660/finalize.go
deitch: Maybe just a warning. Even if we didn't, we should capture the potential error so linters don't complain about it, as `_ = os.RemoveAll(fs.workspace)`

But I think really you are right. Just capture it and warn. At least let the user know that something is lingering.

deitch (changes requested):
Thanks for doing this. I would take your recommendation and at least warn.

filesystem/iso9660/finalize.go
carbonin: > Just capture it and warn

This would be great, but I'm not sure how to accomplish it. We don't have a logger and there's not really a distinction between error types returned typically.
What did you have in mind for this?

deitch: Oh good point. I had forgotten about that. These libraries are not meant to be interacting directly that way. The right thing to do probably is to have various kinds of custom errors, pass those up, and let the caller know that the final state was successful, but that there were secondary errors.

That is far more than I want to think about in this PR, so let's just take your input to date.

deitch (approved):

PR #93, opened 2021-02-04
fix(filesystem/iso9660): fix bug when compare null bytes

t3ls:
"Unable to convert expiration date/time from bytes" was raised when parse some specific iso9660 images, eg:

https://mirrors.tuna.tsinghua.edu.cn/ubuntu-releases/xenial/ubuntu-16.04.7-server-amd64.iso


deitch:
Would you please update the description as to why this is necessary? When does the slice given by `nullBytes` (16 of 48 followed by 0) not match appropriately?

deitch (changes requested):
In addition to the comment in the PR, please:

* create a single var of alternate nullBytes and compare it, rather than building each time, something like: `nullBytesAlt := bytes.Repeat([]byte{0}, len(nullBytes))` and then use that in the `bytes.Compare`
* describe in a comment right above that when it is necessary

PR #97, opened 2021-02-11
Rudimentary Windows sanity check

xaduha:


deitch (changes requested):
I like the idea of a basic sanity check, but I don't think it belongs here this way. I think we should use the matrix capabilities of Github Actions and have a job that does a sanity check across all of the OSes. Something added to ci.yaml like:

```yaml
  sanity:
     name: Smoke Test Sanity Check
    strategy:
      matrix:
        os: [darwin, windows, linux]
     steps:
    - name: checkout
      uses: actions/checkout@v2
    - uses: actions/setup-go@v2
      with:
        go-version: '1.15.7' # The Go version to download (if necessary) and use.
    - run: GOOS=${{ matrix.os }} go build
```     

I might have gotten that a bit wrong, but that is the general idea.

PR #101, opened 2021-02-15
Allow providing a workspace in iso9660.Create

### LINKED ISSUE #100: Make iso workspace base dir configurable
Currently, when building an iso, you have to create the filesystem which creates for you a workspace directory which is always in the system temp dir.

Allowing this to be configurable somehow could solve two issues:
1. I don't want to write to `/tmp`
    - Maybe I have more storage elsewhere or something faster
    - In my case specifically I want all of these writes to go to a volume rather than the container filesystem I'm running on
2. I don't want to double write all my files
    - My use case is: extract an iso to dir `x`, make changes in `x`, recreate an iso from all the files rooted at `x`
    - If I could set the workspace to `x` when creating the filesystem I would save on all the calls I'm making to recreate the whole directory tree in the workspace

As for how to solve this ... it would be tough to do both without making a breaking API change to `iso9660.Create` so I understand if we don't want to do that.
An alternative option is to create a convenience function specifically in the iso9660 package that could create and finalize an iso in one call given the workspace dir (pre-populated with the iso content), the existing parameters to `disk.Create` (or some subset) and the finalize options.

Effectively I would contribute what I have [here](https://github.com/openshift/assisted-service/blob/3bc73c1259d48b711077f5ef7de0c1fde8d066d9/internal/isoutil/isoutil.go#L130-L214) altered for general use.

@deitch what do you think? Is this something that you think would be useful or would you rather add a parameter somewhere for the workspace base dir?

carbonin:
This will allow for users to either create an iso from files in an
existing directory or just to specify a directory to be used as the
scratch space for staging files.

If nothing is provided we will fallback to the previous behavior
which would create a directory in the system temp directory

Fixes #100

### DIFF
--- a/disk/disk.go
+++ b/disk/disk.go
@@ -131,6 +131,7 @@ type FilesystemSpec struct {
 	Partition   int
 	FSType      filesystem.Type
 	VolumeLabel string
+	WorkDir     string
 }
 
 // CreateFilesystem creates a filesystem on a disk image, the equivalent of mkfs.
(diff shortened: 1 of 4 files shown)


deitch (approved):
Looks great. Thank you!

PR #102, opened 2021-03-11
Fix fat32.go test image generator

dwj300:
The change to refactor `CreateFilesystem` broke the ` testdata/fat32.go` helper. This change fixes it.

### DIFF
--- a/filesystem/fat32/testdata/fat32.go
+++ b/filesystem/fat32/testdata/fat32.go
@@ -1,56 +1,58 @@
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


dwj300:
@deitch can you review?

deitch:
This all looks good, but why the reformatting? Didn't gofmt already have this correct?

deitch:
Apparently it did not, as I am checking locally right now.

deitch (approved):

Issue #109, opened 2021-04-09
fat32.Filesystem fails silently when writing files to a directory with many items


RafaelMarinheiro:
I'm trying to write a program that formats an SD card and writes the contents of https://github.com/raspberrypi/firmware/tree/master/boot/ to it. The core of the code is:

```
	d, _ := diskfs.Open(dev)
	sectorSize := 512
	ptable := &mbr.Table{
		LogicalSectorSize:  sectorSize,
		PhysicalSectorSize: sectorSize,
		Partitions: []*mbr.Partition{
			{
				Type:  mbr.Fat32LBA,
				Start: 4 * MiB / uint32(sectorSize),
				Size:  2 * GiB / uint32(sectorSize),
			},
		},
	}
	_ := d.Partition(ptable)
	fs, _ := d.CreateFilesystem(disk.FilesystemSpec{
		Partition: 1, FSType: filesystem.TypeFat32, 
	})
	for _, file := range files {
		dest, _ := fs.OpenFile(...)
		_, _ = dest.Write(file.data)
	}
```

Error handling has been abbreviated as well as directory creation, but both are there in my code.

Once I try to write the files in https://github.com/raspberrypi/firmware/tree/master/boot/overlays (219 entries) to /overlays the code does not report any errors, however only the first ~50-ish files are properly written. Files that are supposedly written later, but to a different directory, are still written correctly.

IMHO this is particularly bad because both `fs.OpenFile` and `dest.Write` fail silently, so the caller/user will just notice this much after the program has executed.

deitch:
Do they not report errors? The above code ignores the errors:

```go
		dest, _ := fs.OpenFile(...)
		_, _ = dest.Write(file.data)
```

Is there a way to reproduce the problem without an SD card? It would be great to have a standalone program that reproduces it.

deitch:
Now that #110 is in, can we see if we can reproduce this?

PR #110, opened 2021-04-12
fix reading of partial clusters in fat32

### LINKED ISSUE #108: fat32.File does not not implement io.Writer correctly.
It seems that multiple calls to `fat32.File.Write()` will simply cause previous writes to be overwritten. This is pretty noticeable if one tries to call io.Copy a large reader (32KB+) into it:

```
var src os.File = io.Open(...)
dest, _ = destfs.OpenFile(path, os.O_CREATE|os.O_RDWR)
written, err := io.Copy(dest, src) 
``` 

In this scenario io.Copy will report all src bytes as written and no error is reported. The final file will contain only the last 32KB of the original file (that's because 32KB is the default buffer size used by io.Copy).

deitch:
Fixes #108

### DIFF
--- a/filesystem/fat32/file.go
+++ b/filesystem/fat32/file.go
@@ -36,7 +36,6 @@ func (fl *File) Read(b []byte) (int, error) {
 	if err != nil {
 		return totalRead, fmt.Errorf("Unable to get list of clusters for file: %v", err)
 	}
-	var lastCluster uint32
 	clusterIndex := 0
 
 	// if there is nothing left to read, just return EOF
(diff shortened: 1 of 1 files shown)


deitch:
@RafaelMarinheiro can you try the branch here and see if it resolves your issue #108? The tests pass now.

RafaelMarinheiro:
Thanks! This fixes the issue with io.Copy, making it work as expected!

Having said that, the code with io.Copy is MUCH slower than the code with ReadAll and a single write. I guess this is happening because now you need to deal with FAT32 bookkeeping on every single Write call. You could have avoided this by postponing some of that bookkeeping to a Close() method in the file, but unfortunately the current API does not have anything like that in it.

Thanks again for the fix!

RafaelMarinheiro:
This seems to have fixed https://github.com/diskfs/go-diskfs/issues/109 as well (at least now I can write a large number of things to the same directory and they show up afterwards). Could it be related in some way?

deitch:
Possibly. The `Read` was broken, which might have done it. Let's merge this in, then follow through on #109

PR #112, opened 2021-04-30
Adds repair capabilities to GPT

thebsdbox:
This adds two public functions to `partition.Table`

`Verify()` which will examine headers and internal data to validate the GPT table to the structure of the disk
`Repair()` which will update the in-memory table so it can be re-written to disk.

### DIFF
--- a/partition/gpt/table.go
+++ b/partition/gpt/table.go
@@ -322,6 +322,75 @@ func (t *Table) toGPTBytes(primary bool) ([]byte, error) {
 	return b, nil
 }
 
+// tableHeaderFromBytes read a partition table from a byte slice, mainly used to validate teh secondary header
+func tableHeaderFromBytes(b []byte, logicalBlockSize, physicalBlockSize int, skipMBR bool) (*Table, error) {
+	// minimum size - gpt entries + header + LBA0 for (protective) MBR
+	minSize := logicalBlockSize
+	if len(b) < minSize {
+		return nil, fmt.Errorf("Data for partition was %d bytes instead of expected minimum %d", len(b), minSize)
+	}
+	gpt := b
+	if skipMBR {
+		gpt = b[logicalBlockSize:]
+	}
+	// start with fixed headers
+	efiSignature := gpt[0:8]
+	efiRevision := gpt[8:12]
+	efiHeaderSize := gpt[12:16]
+	efiHeaderCrcBytes := append(make([]byte, 0, 4), gpt[16:20]...)
+	efiHeaderCrc := binary.LittleEndian.Uint32(efiHeaderCrcBytes)
+	efiZeroes := gpt[20:24]
+	primaryHeader := binary.LittleEndian.Uint64(gpt[24:32])
+	secondaryHeader := binary.LittleEndian.Uint64(gpt[32:40])
+	firstDataSector := binary.LittleEndian.Uint64(gpt[40:48])
+	lastDataSector := binary.LittleEndian.Uint64(gpt[48:56])
+	diskGUID, err := uuid.FromBytes(bytesToUUIDBytes(gpt[56:72]))
+	if err != nil {
+		return nil, fmt.Errorf("Unable to read guid from disk: %v", err)
+	}
+
+	partitionEntryCount := binary.LittleEndian.Uint32(gpt[80:84])
+	partitionEntrySize := binary.LittleEndian.Uint32(gpt[84:88])
+
+	// once we have the header CRC, zero it out
+	copy(gpt[16:20], []byte{0x00, 0x00, 0x00, 0x00})
+	if bytes.Compare(efiSignature, getEfiSignature()) != 0 {
+		return nil, fmt.Errorf("Invalid EFI Signature %v", efiSignature)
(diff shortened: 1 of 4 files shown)


deitch:
This is super interesting. I am going to make some comments to the code. Do you mind explaining here what the scenario is? Why would we need this, and what this does?

partition/gpt/table.go
```diff
@@ -322,6 +322,75 @@ func (t *Table) toGPTBytes(primary bool) ([]byte, error) {
 	return b, nil
 }
 
+// tableHeaderFromBytes read a partition table from a byte slice, mainly used to validate teh secondary header
+func tableHeaderFromBytes(b []byte, logicalBlockSize, physicalBlockSize int, skipMBR bool) (*Table, error) {
```
deitch, line 326: This looks like it is almost completely duplicative of `tableFromBytes()`. Can we wrap these together to avoid code copies?

partition/gpt/table.go
```diff
@@ -525,3 +594,35 @@ func (t *Table) GetPartitions() []part.Partition {
 	}
 	return parts
 }
+
+// Verify will attempt to evaluate the headers and repair mis-aligned
```
deitch, line 598: Is it correct that `Verify()` does not actually repair, but just evaluate? I don't see changes here. Can the comment describe what it actually verifies?

partition/gpt/table.go
```diff
@@ -525,3 +594,35 @@ func (t *Table) GetPartitions() []part.Partition {
...
+	if t.firstDataSector != secondaryTable.firstDataSector {
+		return fmt.Errorf("Error comparing GPT headers expected =>  %d / actual => %d", t.firstDataSector, secondaryTable.firstDataSector)
+
+	}
+	return nil
+}
+
+// Repair will attempt to evaluate the headers fix the header location and re-write the primary and secondary header
+func (t *Table) Repair(diskSize uint64) error {
+	t.secondaryHeader = (diskSize / uint64(t.LogicalSectorSize)) - 1
```
deitch, line 625: Is all `Repair()` doing marking the table as having the secondary header at the last logical second? How would it end up differently?

partition/gpt/table.go
thebsdbox: Possibly.. there was some intertwinned data that would be hard to move around between the functions so I did a small dupe in the first iteration.

partition/gpt/table.go
thebsdbox: Yes, `verify()` is just to perform a few checking tasks.

partition/gpt/table.go
thebsdbox: Repair corrects the table.. We can then call the `write()` function to write a correct pair of GPT tables.

thebsdbox:
> This is super interesting. I am going to make some comments to the code. Do you mind explaining here what the scenario is? Why would we need this, and what this does?

The main scenario here is deploying a partition table on a smaller disk (i.e. 2GB) and then re-writing the contents of that disk (byte-by-byte) to a larger disk (i.e. 4GB). The Linux Kernel, VMware etc.. all complain that the second GPT table is missing and that the first table points to the wrong location.

partition/gpt/table.go
deitch: Is it impossible to just have a single call?

partition/gpt/table.go
deitch: Then I think the comment should be updated.

partition/gpt/table.go
deitch: > corrects the table

Meaning that it corrects the LBA of `secondaryHeader`?

deitch:
> The main scenario here is deploying a partition table on a smaller disk (i.e. 2GB) and then re-writing the contents of that disk (byte-by-byte) to a larger disk (i.e. 4GB). The Linux Kernel, VMware etc.. all complain that the second GPT table is missing and that the first table points to the wrong location.

I will admit, that is not a scenario I had thought of before. So I create a 2GB disk (actual or image), then `dd if=/in/dev of=/out/dev` where `/out/dev` is a disk file or physical disk of size 4GB. When that happens, the first 2GB in `out` are everything, including the secondary header at the last LBA of the 2GB, following which (from just after 2GB until end of 4GB) is just unused data, perhaps 0s or whatever.

In that case, the kernel/VMware/etc look at the disk size and expect the secondary table at the end, no matter what is in the beginning, and so complain?

thebsdbox:
Correct 😀

deitch:
OK, makes sense. Can we address the comments, and then we can run CI and merge in?

adamqqqplay:
@thebsdbox Hi, Thank you very much for your contribution, I wonder if you still have time or interest to complete this PR. It would help us a lot if this feature could be supported, thanks!

adamqqqplay:
> Good question.. the development VM I did all this in is long gone, so I'll need some time to remember what I was doing 😂.
> 
> I'll add it to my todo list.

@thebsdbox Thanks, your work will help us a lot. And I would like to ask, do you have any update with this PR?

Issue #120, opened 2021-06-07
Disk.Table is not populated without partition write


iancoolidge:
It is useful to be able to open a 'Disk' structure for reading, such as in a unit test.

this is not possible currently, as Disk.Table is only populated when writing a partition table to disk, which is a potentially destructive operation.

it should be possible to poulate Disk.Table from a partition read.

deitch:
Yeah, that makes sense. It should go in [here](https://github.com/diskfs/go-diskfs/blob/master/disk/disk.go#L52-L54). Would be happy to take a PR :-)

deitch:
Fixed with #121

PR #121, opened 2021-06-07
disk: When GetPartitionTable succeeds, save result

iancoolidge:
Previously, it is only possible to read a partition after
writing the partition table.

Now, a call to GetPartitionTable will enable calls to GetPartition
and GetFilesystem to succeed without having to write a new partition
table.

### DIFF
--- a/disk/disk.go
+++ b/disk/disk.go
@@ -48,9 +48,16 @@ var (
 
 // GetPartitionTable retrieves a PartitionTable for a Disk
 //
+// If the table is able to be retrieved from the disk, it is saved in the instance.
+//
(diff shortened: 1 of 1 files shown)


deitch (approved):

iancoolidge:
I am a github scrub,

trying to figure out how to associate this with https://github.com/diskfs/go-diskfs/issues/120

deitch:
> trying to figure out how to associate this with #120

Too late, as I merged it, but you just put a line on the PR comment as:

Fixes #120

iancoolidge:
ok -- no problem. Thanks for the quick response Avi.

PR #125, opened 2021-06-18
Add Darwin disk blocks support (/dev/diskX / /dev/rdiskX)

cusspvz:
- Adds support for disk blocks on darwin (`/dev/(r)?disk.+`)
- Updated the /sys/unix dependency and based the `ioctl` syscalls from its constants instead in linux

Notes:
- It still works with regular files on darwin, happens that the previous ioctl syscall flags are specific for some unix systems (linux included), but not for freebsd and darwin.

### DIFF
--- a/diskfs_darwin.go
+++ b/diskfs_darwin.go
@@ -1,5 +1,3 @@
-// +build !darwin linux solaris aix freebsd illumos netbsd openbsd plan9
-
 package diskfs
 
 import (
@@ -9,18 +7,26 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+// this constants should be part of "golang.org/x/sys/unix", but aren't, yet
+const (
+	DKIOCGETBLOCKSIZE         = 0x40046418
+	DKIOCGETPHYSICALBLOCKSIZE = 0x4004644D
+	DKIOCGETBLOCKCOUNT        = 0x40086419
+)
+
 // getSectorSizes get the logical and physical sector sizes for a block device
 func getSectorSizes(f *os.File) (int64, int64, error) {
 	/*
 		ioctl(fd, BLKPBSZGET, &physicalsectsize);
+
 	*/
 	fd := f.Fd()
 
-	logicalSectorSize, err := unix.IoctlGetInt(int(fd), unix.BLKSSZGET)
+	logicalSectorSize, err := unix.IoctlGetInt(int(fd), DKIOCGETBLOCKSIZE)
 	if err != nil {
 		return 0, 0, fmt.Errorf("Unable to get device logical sector size: %v", err)
 	}
-	physicalSectorSize, err := unix.IoctlGetInt(int(fd), unix.BLKPBSZGET)
+	physicalSectorSize, err := unix.IoctlGetInt(int(fd), DKIOCGETPHYSICALBLOCKSIZE)
 	if err != nil {
 		return 0, 0, fmt.Errorf("Unable to get device physical sector size: %v", err)
 	}
--- a/diskfs_unix.go
+++ b/diskfs_unix.go
@@ -15,7 +15,7 @@ func getSectorSizes(f *os.File) (int64, int64, error) {
 		ioctl(fd, BLKPBSZGET, &physicalsectsize);
 	*/
 	fd := f.Fd()
-
+- 
 	logicalSectorSize, err := unix.IoctlGetInt(int(fd), unix.BLKSSZGET)
 	if err != nil {
 		return 0, 0, fmt.Errorf("Unable to get device logical sector size: %v", err)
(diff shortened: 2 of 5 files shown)


deitch:
Thanks for this.

The `+build` statement is the same on both `_linux.go` and `_darwin.go`, which should cause them both to build, which will conflict. I think the build statements need to be adjusted. Sure, they use the `_<GOOS>.go` convention, but the `+build` will conflict with that. In any case, I don't see any reason why this should be only on Linux and Darwin; it can be on other Unix and Unix-like platforms.

I think we need to rename `_linux.go` to `_unix.go` (how it was before), just put `!darwin` in the ``+build` statement of that file, and then have a `_darwin.go` file.

Separately, no need for a `constants_darwin.go` file, if they are only ever used in `diskfs_darwin.go`, we can just put them in there. And we will need comments for the linter.

cusspvz:
@deitch thanks for your review.

I didn't notice the `+build` statement, totally missed it.

In the meantime, I've applied the changes you've requested. Let me know if it looks good to merge.

deitch:
I am going to ask that you make 2 more changes.

1. Squash your commits to a single one.
2. Keep the original values that were replaced with the `unix.*` constants in the files but commented out. I have had headaches with missing things in `sys/x/unix` in the past, or conflicts across OSes. I want to keep it that so that we have a reference and save days of tears in the future.

Thanks!

mepholic:
+1 on this, as it duplicates the fix I made in #128

cusspvz:
@mepholic thanks for the comment on this, this got stalled because I forgot to do these changes... :/ 

@deitch done! 🙌🏼

diskfs_darwin.go
```diff
@@ -0,0 +1,34 @@
...
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
+// this constants should be part of "golang.org/x/sys/unix", but aren't, yet
+const (
+	DKIOCGETBLOCKSIZE         = 0x40046418
+	DKIOCGETPHYSICALBLOCKSIZE = 0x4004644D
+	DKIOCGETBLOCKCOUNT        = 0x40086419
```
deitch, line 14: Is this `DKIOCGETBLOCKCOUNT` used anywhere?

diskfs_darwin.go
```diff
@@ -0,0 +1,34 @@
...
+	"fmt"
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
+// this constants should be part of "golang.org/x/sys/unix", but aren't, yet
+const (
+	DKIOCGETBLOCKSIZE         = 0x40046418
+	DKIOCGETPHYSICALBLOCKSIZE = 0x4004644D
```
deitch, line 13: Are these correct? The values are different than the replaced constants, and I am pretty sure (not definitely) that those values were correct for Darwin.

diskfs_unix.go
```diff
@@ -13,14 +13,14 @@ import (
 func getSectorSizes(f *os.File) (int64, int64, error) {
 	/*
 		ioctl(fd, BLKPBSZGET, &physicalsectsize);
-
 	*/
 	fd := f.Fd()
-	logicalSectorSize, err := unix.IoctlGetInt(int(fd), blksszGet)
+
```
deitch, line 18: Looks like you added an extra line here.

diskfs_darwin.go
cusspvz: they were not. the unix values work on darwin if we're using an image file instead of a block but doesn't work for a `rdisk` block. The values I'm posting works for both cases.

diskfs_darwin.go
cusspvz: Nope, but in case we need that, I though would be better to let it defined

diskfs_unix.go
cusspvz:
```suggestion
- 
```

cusspvz:
@deitch you can squash the commits on merge with Github, right?

diskfs_darwin.go
deitch: OK, makes sense. It is a headache to track those down.

diskfs_darwin.go
deitch: 👍

deitch:
CI is unhappy, looks like `go vet` failed.

cusspvz:
Do you have a clue on why that happened? It runs fine on my end, and the code is basically the same as before. 🤔

deitch:
Are you running on a Mac or Windows? Try `GOOS=linux go vet ./...` to force it to vet for linux?

amgedr:
Seems like a typo in diskfs_unix.go.

deitch:
@cusspvz where does this stand?

cusspvz:
@deitch @amgedr fixed the typo at `diskfs_unix.go`. 🙌🏼

deitch (approved):

PR #126, opened 2021-07-27
Check the filepath.WalkFunc error in iso9660.Finalize

carbonin:
If there is an error getting the file information at any point in the
tree info will be nil and err will be set. The WalkFunc must decide how
to handle the passed error.

In this case, as we can't finalize the iso, we choose to error out.

Before this fix, a failure to access some file in the tree would cause a
panic.

### DIFF
--- a/filesystem/iso9660/finalize.go
+++ b/filesystem/iso9660/finalize.go
@@ -824,7 +824,10 @@ func walkTree(workspace string) ([]*finalizeFileInfo, map[string]*finalizeFileIn
 	dirList := make(map[string]*finalizeFileInfo)
 	fileList := make([]*finalizeFileInfo, 0)
 	var entry *finalizeFileInfo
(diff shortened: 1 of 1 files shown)


deitch (approved):
Any odd side effects, such as files that should be skipped even if they give an error?

Eh, it doesn't matter. If that _does_ exist, it should be handled explicitly. Let's enable it and catch those as needed.

carbonin:
> Any odd side effects, such as files that should be skipped even if they give an error?

I think if some cases were resulting in an error we'd have seen a panic from [this line](https://github.com/diskfs/go-diskfs/pull/126/files#diff-b3e82302abf33b4305a7e0c7ecc980b3260123874801c3a622ad89d4a8612e16R832) as generally the file info would be `nil` in an error case. At least that's what I saw.

PR #128, opened 2021-08-29
Use sys/unix constants for ioctls

mepholic:
This fixes functionality on multiple CPU architectures.

The previously hardcoded blksszGet and blkpbszGet values are the correct
ioctl values on only a few architectures. PowerPC, MIPS, and Sparc use
different bit layouts in their ioctls, so it's best to get these values
from the sys/unix library.

### DIFF
--- a/diskfs.go
+++ b/diskfs.go
@@ -118,8 +118,6 @@ import (
 //    so we use the default sector size of 512, per Rod Smith
 const (
 	defaultBlocksize, firstblock int = 512, 2048
-	blksszGet                        = 0x1268
-	blkpbszGet                       = 0x127b
 )
 
 // Format represents the format of the disk
--- a/diskfs_unix.go
+++ b/diskfs_unix.go
@@ -16,11 +16,11 @@ func getSectorSizes(f *os.File) (int64, int64, error) {
 
 	*/
 	fd := f.Fd()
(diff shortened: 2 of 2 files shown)


deitch:
Thanks @mepholic 

Actually, we originally had that, but had to hard-code it because it isn't available on every platform, notably Darwin. I have no idea why, but there it is.

If you can remove the `darwin` from the `// +build` line, and then create a separate `diskfs_darwin.go` with `// +build darwin` which contains the hard-coded, then I think we will be covered.

deitch (changes requested):
As discussed in the comment, please remove `darwin` from building on `diskfs_unix.go`, create a separate `diskfs_darwin.go` and we should be good.

Alternatively, we could have a `diskfs_unix.go` that is for all, and move the constants `blksszGet` and `blkpbszGet` out of `diskfs.go` and into `constants_unix.go` and `constants_darwin.go`, leaving `diskfs_unix.go` unchanged.

I think I like the second option better - fewer changes, much less repetition - but whatever works.

yadutaf:
It looks like a variant of this PR was merged in the meantime. Indeed, the constants are used in `diskfs_unix.go`. Maybe this can be closed ?

deitch:
I think so. If it is needed, we always can reopen

Issue #130, opened 2021-10-01
FAT32 issues with io.Copy


abarisani:
Hello there.

I am evaluating go-diskfs (nice project!) for use with our [tamago](https://github.com/f-secure-foundry/tamago) framework to port [interlock](https://github.com/f-secure-foundry/interlock) to bare metal. The integration is promising however I am stumbling upon some issues.

First of all while using `io.Copy` on a FAT32 file opened with `OpenFile` I notice that performance is pretty slow, it seems that write speed decreases as the file gets larger, I am using something similar to the following:

```
f, _ := storage.OpenFile(osPath, os.O_RDWR|os.O_CREATE|os.O_EXCL|os.O_TRUNC)
io.Copy(f, buf)
```

I think this relates to https://github.com/diskfs/go-diskfs/pull/110#issuecomment-818249267

Additionally I notice that writes to the underlying FileSystem can be of sizes which are not a multiple of its block size, I am not sure if this is intended or not.

Despite this being correct or not, for sure it's problematic that `WriteAt` errors are not all trapped, for example my custom block device (which reads/writes from an SD card) does not allow writes which are not a multiple of block size, the `WriteAt` error was raised but it's never caught in `fat32.allocateSpace` (also not in some instances of `fat32.Create`).

So my issues are the following:

  * can write performance be improved ?
  * is it correct for `WriteAt` to be invoked without honoring the `FileSystem` block size ?
  * all `WriteAt` errors should be trapped.

Thanks!

deitch:
Hi @abarisani 

> nice project!

Thanks! It is fully OSS without corporate backing, so push it ahead whenever possible.

> I notice that performance is pretty slow, it seems that write speed decreases as the file gets larger, I am using something similar to the following:

That is possible. Do you have any benchmarks? It could be the processing, it could be the read from the file, it could be the underlying filesystem (you said you were opening a FAT32 file, and not a raw block device).

> Additionally I notice that writes to the underlying FileSystem can be of sizes which are not a multiple of its block size, I am not sure if this is intended or not.

Do you mean the blocksize of the filesystem on the file image? Or the blocksize of the underlying filesystem on which the image exists? I assume you mean the former, as the latter would be out of scope and handled by the OS on which it is running.

> all WriteAt errors should be trapped.

Agreed. Care to submit a PR?

abarisani:
[...]

PR #131, opened 2021-10-01
trap FAT32 disk WriteAt errors

abarisani:
Fix for untrapped errors raised by FAT32 disk (mentioned in https://github.com/diskfs/go-diskfs/issues/130).

### DIFF
--- a/filesystem/fat32/file.go
+++ b/filesystem/fat32/file.go
@@ -146,7 +146,10 @@ func (fl *File) Write(p []byte) (int, error) {
 			if toWrite > int64(len(p)) {
 				toWrite = int64(len(p))
 			}
-			file.WriteAt(p[0:toWrite], int64(offset)+fs.start)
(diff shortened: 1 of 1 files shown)


deitch (approved):
I prefer the idiomatic

```go
			if _ , err := file.WriteAt(p[0:toWrite], int64(offset)+fs.start); err != nil {
 				return totalWritten, fmt.Errorf("Unable to write to file: %v", err)
 			}
```

But I wouldn't hold it up for that.

abarisani:
> I prefer the idiomatic
> 
> ```go
> 			if _ , err := file.WriteAt(p[0:toWrite], int64(offset)+fs.start); err != nil {
>  				return totalWritten, fmt.Errorf("Unable to write to file: %v", err)
>  			}
> ```
> 
> But I wouldn't hold it up for that.

Me too! But I forced myself to use the existing style within the code ;). Feel free to change it as you please.

PR #135, opened 2021-11-03
Add the ability to adjust the sector size while creating Disks

### LINKED ISSUE #133: support creating 4K raw images via Create (or new function)?
Hi,

We have a bunch of code in snapd / Ubuntu Core around creating partitions / structures that now via the ubuntu-image rewrite uses this library (see https://github.com/canonical/ubuntu-image for details, specifically `makeDisk()` in common_states.go) to create a raw .img file for a volume that then gets flashed to a device (or in the case of our tests used in a VM)

It would be great both for creating images for consumption on physical media that has 4K physical sector size / block size and for our own VM based testing to be able to create disk images via some option that use 4K as the sector size instead of the default 512 that is currently hard-coded here in diskfs.go.

Maybe a new function CreateWithSectorSize or some additional Format flag in Create()? 

Thanks,
Ian

sil2100:
This fixes #133 .

As discussed on the issue above, per recommendation I added an additional parameter to the `diskfs.Create()` function that allows customization of sector size - currently only in the case of regular files.

Questions here:
1. Should we also override the sector size for block devices in this case? I wasn't sure so I left it to auto-detect always.
2. Naming. It's hard for me to find good names for the values!

Please be gentle, Go is still not very native for me.

### DIFF
--- a/diskfs.go
+++ b/diskfs.go
@@ -17,7 +17,7 @@
 //     size := 10*1024*1024 // 10 MB
 //
 //     diskImg := "/tmp/disk.img"
-//     disk := diskfs.Create(diskImg, size, diskfs.Raw)
+//     disk := diskfs.Create(diskImg, size, diskfs.Raw, diskfs.SectorSizeDefault)
 //
 //     fs, err := disk.CreateFilesystem(0, diskfs.TypeFat32)
 //
@@ -29,7 +29,7 @@
 //     diskSize := 10*1024*1024 // 10 MB
 //
 //     diskImg := "/tmp/disk.img"
-//     disk := diskfs.Create(diskImg, size, diskfs.Raw)
+//     disk := diskfs.Create(diskImg, size, diskfs.Raw, diskfs.SectorSizeDefault)
 //
 //     table := &mbr.Table{
 //       LogicalSectorSize:  512,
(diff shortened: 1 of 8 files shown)


diskfs.go
```diff
@@ -168,16 +180,24 @@ func writableMode(mode OpenModeOption) bool {
...
 		size          int64
-		lblksize      = int64(defaultBlocksize)
-		pblksize      = int64(defaultBlocksize)
+		lblksize      int64
+		pblksize      int64
 		defaultBlocks = true
 	)
 	log.Debug("initDisk(): start")
 
+	if sectorSize != SectorSizeDefault {
```
deitch, line 193: This kind of if/else is a bit non-idiomatic in go. Normally you would keep the default setting, and override if needed. Sort of:

```go
var (
		lblksize      = int64(defaultBlocksize)
 		pblksize      = int64(defaultBlocksize)
)

// ...

// change it if it is different
if lblksize != sectorSize {
		lblksize = int64(sectorSize)
 		pblksize = int64(sectorSize)
}
```

deitch (changes requested):
Just some stylistic go stuff.

sil2100:
Thanks! Adjusted!

deitch (approved):

deitch:
Thanks @sil2100

PR #136, opened 2021-11-09
Support passing sector size to Open*() calls as well.

sil2100:
Forgot that we also need to support setting sector size for `Open*()` calls, since when we open file images there is no way of determining the sector size that got used - and this might make a difference.

This is a follow up to #135 . Sorry for not bundling that in, I was too focused on my `Create()` use-case!

### DIFF
--- a/diskfs.go
+++ b/diskfs.go
@@ -117,9 +117,9 @@ import (
 // when we use a disk image with a GPT, we cannot get the logical sector size from the disk via the kernel
 //    so we use the default sector size of 512, per Rod Smith
 const (
-	defaultBlocksize, firstblock int = 512, 2048
-	blksszGet                        = 0x1268
-	blkpbszGet                       = 0x127b
+	defaultBlocksize = 512
+	blksszGet        = 0x1268
+	blkpbszGet       = 0x127b
 )
 
 // Format represents the format of the disk
@@ -271,7 +271,7 @@ func checkDevice(device string) error {
 // Open a Disk from a path to a device in read-write exclusive mode
 // Should pass a path to a block device e.g. /dev/sda or a path to a file /tmp/foo.img
 // The provided device must exist at the time you call Open()
-func Open(device string) (*disk.Disk, error) {
+func Open(device string, sectorSize SectorSize) (*disk.Disk, error) {
 	err := checkDevice(device)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


diskfs.go
```diff
@@ -117,9 +117,9 @@ import (
 // when we use a disk image with a GPT, we cannot get the logical sector size from the disk via the kernel
 //    so we use the default sector size of 512, per Rod Smith
 const (
-	defaultBlocksize, firstblock int = 512, 2048
```
deitch, line 120: Is `firstblock` not used anywhere at all?

deitch (commented):
I had one question about `firstblock`, which surprised me.

I don't object to passing sector size, but it places an additional burden on the user. 

1. In block device cases, the sector size should be figured out
2. In file cases, what if I just want to use the default? That isn't handled here, and why should I think about it? `diskfs.Open("/path/to/file")` is easy to understand; `diskfs.Open("/path/to/file", sectorSize)`. Size? What size do I need? Isn't there just a default?

I am wondering if we shouldn't collapse the two `Open` and `OpenWithMode` to a single `Open`, which accepts variadic functional arguments.

diskfs.go
sil2100: Yes, that is my understanding! I grepped for `firstblock` and it was not used anywhere, and since it's lowercase it seems to be an internal symbol as well. Not sure why it's there! So I decided to remove it as it was confusing.

sil2100:
Thank you for the review! There is a default value to use: `diskfs.SectorSizeDefault`, which is basically 0 value but translates in `initDisk()` to the default behavior, e.g. just taking the default values. For block devices, well, as mentioned in my previous PR description I decided that the sector size override will only be doing stuff for non-block devices. So if you try to `Open()` a block device, it'll just try fetching the actual sector size of the device and use that.
This can be changed of course, but depends if we want it.

As for variadic parameters: I would be fine with that! I leave this decision up to you core diskfs developers! But I agree, this does break stuff and require spaghetti coding the default value everywhere. In Python I'd just make this as an optional argument from the start - so maybe doing it as a vararg is better.
Should I try that then?

diskfs.go
deitch: go figure. Thanks for finding it. I am surprised lint didn't catch it.

deitch:
Check out #137

sil2100:
This is awesome, thanks! Much cleaner and 'slick'. Closing this one in that case.

PR #140, opened 2022-02-28
Adding hyphen to the regex because it is a valid character in file names

phillipsj:
This adds the hyphen to the regex and adds unit tests.

Signed-off-by: Jamie Phillips <jamie.phillips@suse.com>

### DIFF
--- a/filesystem/iso9660/finalize.go
+++ b/filesystem/iso9660/finalize.go
@@ -888,7 +888,7 @@ func calculateShortnameExtension(name string) (string, string) {
 	extension = strings.ToUpper(extension)
 
 	// replace illegal characters in shortname and extension with _
-	re := regexp.MustCompile("[^A-Z0-9_]")
+	re := regexp.MustCompile("[^A-Z0-9_-]")
 	shortname = re.ReplaceAllString(shortname, "_")
 	extension = re.ReplaceAllString(extension, "_")
 
--- a/filesystem/iso9660/finalize_internal_test.go
+++ b/filesystem/iso9660/finalize_internal_test.go
@@ -200,3 +200,30 @@ func TestCollapseAndSortChildren(t *testing.T) {
 		t.Log(output)
 	}
 }
+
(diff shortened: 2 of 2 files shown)


phillipsj:
@deitch

deitch:
@phillipsj thanks for bringing this in. I am not sure it is correct. The ecma-119 spec says the following:

```
7.4 Character sets and coding 7.4.1 d-characters and a-characters

Except as specified in 7.4.4, the characters in the descriptors shall be coded according to ECMA-6 (see Appendix A).
The 37 characters in the following positions of the International Reference Version are referred to as d-characters:
3/0 to 3/9
4/1 to 5/10
5/15

7.5 File Identifier
7.5.1 File Identifier format
A File Identifier shall consist of the following sequence:
- File Name: A sequence of zero or more d-characters or d1-characters;
- SEPARATOR 1;
- File Name Extension: A sequence of zero or more d-characters or d1-characters;
- SEPARATOR 2;
- File Version Number: Digits representing a number from 1 to 32 767.
```

So the file name and extension can be just d-characters (ecma-119) or d1-characters (iso-9660), which basically are the same. The d-character list is as above.

The ecma-6 list is [here](https://www.ecma-international.org/wp-content/uploads/ECMA-6_6th_edition_december_1991.pdf).

* 3/0 to 3/9 - characters 0-9
* 4/1 to 5/10 - capital A to capital Z
* 5/15 - `_`

I don't see `-` in there anywhere.

phillipsj:
Hmm, interesting that so many things use that.

phillipsj:
@deitch it seems that is a Joliet thing, we are working with Windows. Would you be open to adding Joliet extension support? https://docs.microsoft.com/en-us/windows/win32/imapi/disc-formats#joliet

deitch:
Sure, it has been a "todo" item for a while, see [here](https://github.com/diskfs/go-diskfs/blob/274de576a1a50620d33bc7ce84064e891780dab2/README.md#tests).

It always is a question of time to do it.

PR #144, opened 2022-05-06
allow gpt partitions of 1 block

therjak:
This fixes 2 off by one issues in creating GPT partitions.

As EndingLBA is the last included LBA of a partition Start and End can point to the same LBA.
The logic "part.End = start + size/uint64(blocksize) - 1" in initEntry also applies to WriteContents.

I also added checks to ensure partition sizes are a multiple of the logicalSectorSize.  Other sizes are not possible by the design of GPT partition entries. The alignment should be the responsibility of the caller but as most probably use sizes in full MiB violations seem rare.


deitch (approved):
As long as CI is green, this looks good to go. Thanks for submitting!

PR #152, opened 2022-09-29
iso9660/finalize: parse empty path correctly

Foxboron:
findEntry does not understand what to do when path.Dir returns "." on a file with no directory. This happens when we don't pass a BootCatalog, the default ones do not contain a path, or when we include a BootFile without any path.

Signed-off-by: Morten Linderud <morten.linderud@mullvad.net>

### DIFF
--- a/filesystem/iso9660/finalize.go
+++ b/filesystem/iso9660/finalize.go
@@ -296,7 +296,7 @@ func (fi *finalizeFileInfo) findEntry(p string) (*finalizeFileInfo, error) {
 		err    error
 	)
 	parts := splitPath(p)
-	if len(parts) == 0 {
+	if len(parts) == 0 || p == "." {
 		target = fi
 	} else {
 		current := parts[0]


Foxboron:
If anyone needs help to avoid this bug you just need to ensure there is always a path prefixed the files we give the library.

<pre>
options := iso9660.FinalizeOptions{
    VolumeIdentifier: "testISO",
    ElTorito: &iso9660.ElTorito{
        BootCatalog: "/BOOT.CAT",
        Entries: []*iso9660.ElToritoEntry{
            {
                Platform:  iso9660.EFI,
                Emulation: iso9660.NoEmulation,
                BootFile:  "/something.efi",
            },
        },
    },
}
</pre>

deitch:
@Foxboron thanks for this. I have no objection to the logic that says, "if path has no parts or is `.`, then it is the target", i.e. adding the "or is `.`". But I couldn't figure out how you triggered it. Anything other than `""` should return some string, and anything with `""` should have zero parts. So how do you actually trigger it?

Foxboron:
<pre>
options := iso9660.FinalizeOptions{
    VolumeIdentifier: "testISO",
    ElTorito: &iso9660.ElTorito{
        BootCatalog: "BOOT.CAT",
        Entries: []*iso9660.ElToritoEntry{
            {
                Platform:  iso9660.EFI,
                Emulation: iso9660.NoEmulation,
                BootFile:  "something.efi",
            },
        },
    },
}
</pre>

The above triggers it quite easily. The issue is that `path.Dir("something.efi")` will return "." which the code doesn't understand.

deitch:
OK, I see it.

CI is failing because of linter errors. They have just been fixed. Please rebase on latest, let's get CI green, and then we can merge in.

Foxboron:
I have rebased this now :)

deitch (approved):

PR #158, opened 2022-12-01
separate reading of GPT table and partition entries

### LINKED ISSUE #157: gpt can have size larger than minimum size
This was raised by @carbonin in #156 

To recreate the problem:

1. Download [fedora.iso](https://download.fedoraproject.org/pub/fedora/linux/releases/37/Server/x86_64/iso/Fedora-Server-dvd-x86_64-37-1.7.iso)
2. Run the sample [iso_info.go](https://github.com/diskfs/go-diskfs/blob/master/examples/iso_info.go) program on it
3. See the `"slice bounds out of range [:32768] with capacity 17408"` error

The root of this problem is the following.

When we call [tableFromBytes](https://github.com/diskfs/go-diskfs/blob/d704acd18db2dae47d0cbb2cbed5850016b13fc5/partition/mbr/table.go#L83), we pass it a `[]byte` whose size is the expected size of the GPT table. That size is calculated [here](https://github.com/diskfs/go-diskfs/blob/d704acd18db2dae47d0cbb2cbed5850016b13fc5/partition/gpt/table.go#L506), where we say the size is:

```go
gptSize+logicalBlockSize*2
```

where

```go
// gptSize max potential size for partition array reserved 16384
gptSize                  = 128 * 128
```

so we allow 16384 + 2 blocks of 512 = 17408. 

The error comes [here](https://github.com/diskfs/go-diskfs/blob/d704acd18db2dae47d0cbb2cbed5850016b13fc5/partition/gpt/table.go#L374-L376):

```go
	partArrayStart := partitionEntryFirstLBA * uint64(logicalBlockSize)
	partArrayEnd := partArrayStart + uint64(partitionEntryCount*partitionEntrySize)
	bpart := b[partArrayStart:partArrayEnd]
```

In the case of the `fedora.iso`:

* `partitionEntryFirstLBA=2`
* `logicalBlockSize=512`
* `partitionEntryCount=248`
* `partitionEntrySize=128`

The problem appears to be that we are assuming a `gptSize` of 128 entries, when this iso is saying it has 248. We are reading that count correctly. I suspect that we are taking the minimum size, rather than actual size, and need a mechanism to read the beginning of the table, get basic header information, then reread to get the partition information.

deitch:
Fixes #157 

When we read the GPT table, we read the bytes necessary for table including assumed partition entries, then try to parse them all.

This doesn't work, because our assumptions of the partition entries could very well be wrong.

This PR changes it to a few steps:

1. Read the bytes just for the GPT header
2. Parse and validate those
3. Using the data from the header, read the bytes for the partition entries
4. Validate the checksum on the bytes for the partition entries
5. Parse the partition entries and add to the table struct

### DIFF
--- a/partition/gpt/table.go
+++ b/partition/gpt/table.go
@@ -14,7 +14,6 @@ import (
 
 // gptSize max potential size for partition array reserved 16384
 const (
-	gptSize                  = 128 * 128
 	mbrPartitionEntriesStart = 446
 	mbrPartitionEntriesCount = 4
 	mbrpartitionEntrySize    = 16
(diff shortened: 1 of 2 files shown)


deitch:
@carbonin see here.

PR #160, opened 2022-12-02
Stricter Fat32 volume validation, fix Fat32 MkDir when volume name matches new directory name

apires:
This patch adds a few things:
 - Validates volume name a bit closer to the fat spec
 - (bug) When we're walking a directory for mkdir we compare what we're looking for by directory entry nodes, rather than on a file/directory basis. This leaves the root volume entry as an unhandled case, specially if the volume name matches the name of the file we're trying to create. This patchset skips the the root directory node in this traversal.

### DIFF
--- a/filesystem/fat32/directory.go
+++ b/filesystem/fat32/directory.go
@@ -1,6 +1,7 @@
 package fat32
 
 import (
+	"fmt"
 	"time"
 )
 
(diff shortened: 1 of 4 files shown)


deitch (approved):
Excellent, thank you.

deitch:
Some lint errors to clean up.

PR #164, opened 2023-01-04
Improved error handling for Create()

shellwhale:
Some errors returned by the Create function were too generic. 

Here is an example of a previous error output : 

`could not create device my-disk.iso`

Now it's as follows : 

`could not create device my-disk.iso: The file exists.`

 I changed the return statements in order to add the unwrapped original syscall error.

### DIFF
--- a/diskfs.go
+++ b/diskfs.go
@@ -344,11 +344,11 @@ func Create(device string, size int64, format Format, sectorSize SectorSize) (*d
 	}
 	f, err := os.OpenFile(device, os.O_RDWR|os.O_EXCL|os.O_CREATE, 0o666)
 	if err != nil {
-		return nil, fmt.Errorf("could not create device %s", device)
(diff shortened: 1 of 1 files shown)


deitch:
Is it better to unwrap, or just wrap it? E.g. what if instead of:

```go
		return nil, fmt.Errorf("could not create device %s: %v", device, errors.Unwrap(err))
```

we did:

```go
		return nil, fmt.Errorf("could not create device %s: %w", device, err)
```

Just include the original error but wrap it?

shellwhale:
Well if we don't `Unwrap` we get :
`could not create device my-disk.iso: open my-disk.iso: The file exists.`

instead of 
`could not create device my-disk.iso: The file exists.`

My perspective is that it's irrelevant to show that open is being used in the background, from a diskfs user viewpoint, 
To me Unwrapping the error makes it simpler for the user.

deitch (approved):

PR #168, opened 2023-02-09
examples: Update dependencies to use latest go-diskfs

adamqqqplay:
We need to update the version of dependencies in the example in time to help users use this project better.

Signed-off-by: Qinqi Qu <quqinqi@linux.alibaba.com>


deitch (approved):

deitch:
Thanks!

PR #170, opened 2023-02-13
build: Update module dependencies to latest

adamqqqplay:
This patch updates all dependencies to the latest versions to avoid potential bugs.

Signed-off-by: Qinqi Qu <quqinqi@linux.alibaba.com>


deitch (approved):
As long as it works...

PR #173, opened 2023-03-08
Detect block device size using platform-specific IOCTL on Mac & Linux

yadutaf:
Currently, diskfs attempts to detect the size of a block device using a
"seek" based approach. Unfortunately, while this looks generic, it does
not work on Mac, at least not in a VM.

Based on the previous commit, this commit adds a Mac and Linux
implementation for getting the size of a Block device through IOCTL
inspired by https://github.com/tytso/e2fsprogs/blob/master/lib/ext2fs/getsize.c

This was manually tested using a Linux loopback device and in a Mac
virtual machine.

### DIFF
--- a/diskfs.go
+++ b/diskfs.go
@@ -105,7 +105,6 @@ package diskfs
 import (
 	"errors"
 	"fmt"
-	"io"
 	"os"
 
 	log "github.com/sirupsen/logrus"
@@ -213,14 +212,9 @@ func initDisk(f *os.File, openMode OpenModeOption, sectorSize SectorSize) (*disk
 	case mode&os.ModeDevice != 0:
 		log.Debug("initDisk(): block device")
 		diskType = disk.Device
-		file, err := os.Open(f.Name())
+		size, err = getBlockDeviceSize(f)
 		if err != nil {
-			return nil, fmt.Errorf("error opening block device %s: %s", f.Name(), err)
-		}
-		defer file.Close()
-		size, err = file.Seek(0, io.SeekEnd)
-		if err != nil {
-			return nil, fmt.Errorf("error seeking to end of block device %s: %s", f.Name(), err)
+			return nil, fmt.Errorf("error getting block device %s size: %s", f.Name(), err)
 		}
 		lblksize, pblksize, err = getSectorSizes(f)
 		log.Debugf("initDisk(): logical block size %d, physical block size %d", lblksize, pblksize)
--- a/diskfs_darwin.go
+++ b/diskfs_darwin.go
@@ -14,6 +14,22 @@ const (
 	DKIOCGETBLOCKCOUNT        = 0x40086419
 )
 
+// getBlockDeviceSize get the size of an opened block device in Bytes.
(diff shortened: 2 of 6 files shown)


yadutaf:
Let me know if you'd rather see 2 distinct PRs (one for the refactoring and 1 for the IOCTL)

deitch:
I like this, no need to split it. I have some questions, going to comment inline.

diskfs.go
```diff
@@ -213,17 +212,10 @@ func initDisk(f *os.File, openMode OpenModeOption, sectorSize SectorSize) (*disk
...
 		if err != nil {
-			return nil, fmt.Errorf("error opening block device %s: %s", f.Name(), err)
+			return nil, fmt.Errorf("error getting block device %s size: %s", f.Name(), err)
 		}
-		defer file.Close()
-		size, err = file.Seek(0, io.SeekEnd)
-		if err != nil {
-			return nil, fmt.Errorf("error seeking to end of block device %s: %s", f.Name(), err)
-		}
-		lblksize, pblksize, err = getSectorSizes(f)
```
deitch, line 225: So how are logical block size and physical block size (`lblksize` and `pblksize`, respectively) supposed to be set?

diskfs_darwin.go
```diff
@@ -14,13 +14,24 @@ const (
...
+	return int64(blockSize) * int64(blockCount), nil
+}
+
 // getSectorSizes get the logical and physical sector sizes for a block device
 func getSectorSizes(f *os.File) (logicalSectorSize, physicalSectorSize int64, err error) {
-	//nolint:gocritic // we keep this for reference to the underlying syscall
-	/*
-		ioctl(fd, BLKPBSZGET, &physicalsectsize);
-
-	*/
```
deitch, line 23: Why remove this comment?

diskfs_other.go
```diff
@@ -0,0 +1,19 @@
+//go:build !windows && !linux && !darwin
+// +build !windows,!linux,!darwin
```
deitch, line 2: The `go.mod` now has go 1.19; I don't think you need the obsolete `// +build` line anymore

diskfs_darwin.go
yadutaf: Because it looks strange to keep a reference to a Linux ioctl in Darwin code.

diskfs.go
yadutaf: It looks very much like a big mistake 🤔. It also means that my manual tests were not relevant…

diskfs_other.go
yadutaf: Removed. Thank.

diskfs.go
yadutaf: Fixed and re-tested on Mac VM & Linux loopback

diskfs_darwin.go
deitch: Fair enough.

deitch (approved):

deitch:
Thank you!

PR #174, opened 2023-03-24
Replace os.Chdir logic in ISO9660 with absolute paths to allow for concurrent use

borisroman:
Hello,

I have a use case where I create multiple ISO files concurrently with different contents and names.

The `os.Chdir()` call changes the working dir for the entire app. Having multiple go routines that call the function does end up in a race condition which has the correct working dir.

This patch uses absolute paths and does not rely on the working directory to be changed.

Thanks for maintaining the package!

### DIFF
--- a/filesystem/iso9660/finalize.go
+++ b/filesystem/iso9660/finalize.go
@@ -400,7 +400,7 @@ func (fs *FileSystem) Finalize(options FinalizeOptions) error {
 	}
 
 	// starting point
-	root := dirList["."]
+	root := dirList["/"]
 	root.addProperties(1)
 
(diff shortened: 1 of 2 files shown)


deitch:
Yeah, that `os.Chdir()` never was great.

`fs.FS` has made it into go since this was written. Maybe redo the whole this with that?

```go
	fsys := os.DirFS(workspace)
 	dirList := make(map[string]*finalizeFileInfo)
 	fileList := make([]*finalizeFileInfo, 0)

 	var entry *finalizeFileInfo
 	err = fs.WalkDir(fsys, ".", func(fp string, d fs.DirEntry, err error) error {
```

Probably requires a bit more cleanup than that, but it would solve your problem with even less effort.

PR #175, opened 2023-04-03
WIP: fs.FS support

### LINKED ISSUE #169: Implement standard io.FS interface and friends
I've stumbled upon this interesting package during my research on how to serve contents of iso file over http.

Definitely, I could loop-mount it and just serve static files from mount point, but the idea of implementing it completely in go took me over.

So, my web part is implemented using gin-gonik and it has a very handy [FileFromFS](https://github.com/gin-gonic/gin/blob/master/context.go#L1043) method, that operates through standard io.FS interface.

I'm quite a new gopher and do not have enough knowlegde to improve such a big go projects, so I've started with a wrapper around go-diskfs. It, actually work, but, it's very much suboptimal in many ways as it does not have access to the internals of this project.

If this would be implemented, the actual code of serving files from iso will boil down to the following:

```go
func main() {
        app := gin.Default()
	
	disk, err := diskfs.Open("/path/to/image.iso")
	if err != nil {
		log.Fatalf("failed to open file: %s", err)
	}
	
	isofs, err := disk.GetFilesystem(0)

	if err != nil {
		log.Fatalf("failed to get partition 0: %s", err)
	}
	
	app.GET("/iso/*path", func(c *gin.Context) {
		c.FileFromFS(c.Param("path"), http.FS(isofs))
	})
	
	a.Run(":8080")
}
```
Any help will be appreciated!

If you're interested in the idea behind this - it's quite trivial: to serve linux distribution repos from the iso file over http for group access.

aol-nnov:
@deitch please find my first effors of implementing fs.FS here. Only iso9660 is supported as yet.

For me, there it is a dilemma: either to break the api and implement whole fs.FS support or implement wrapper functions. Wrappers look clumsily and, either way, require modifications of existing functions...

As for newly created `diskfs.OpenFile`, I've made it during my efforts of fiddling with "multi-layered fs.FS", i.e. say, I have a `DirFS`, where I call fs.Open yeiding an `fs.File`, which I pass to `diskfs.OpenFile` to get to the next layer. Again, it's not very handy, as `diskfs.initDisk` requires `os.File` - another "why" in my journey..

closes #169

### DIFF
--- a/diskfs.go
+++ b/diskfs.go
@@ -327,6 +327,32 @@ func Open(device string, opts ...OpenOpt) (*disk.Disk, error) {
 	return initDisk(f, ReadWriteExclusive, opt.sectorSize)
 }
 
+func OpenFile(f *os.File, opts ...OpenOpt) (*disk.Disk, error) {
+	// err := checkDevice(device)
+	// if err != nil {
+	// 	return nil, err
+	// }
+
+	opt := openOptsDefaults()
+	for _, o := range opts {
+		if err := o(opt); err != nil {
+			return nil, err
+		}
+	}
+
+	// m, ok := openModeOptions[opt.mode]
+	// if !ok {
+	// 	return nil, errors.New("unsupported file open mode")
+	// }
+
+	// f, err := os.OpenFile(device, m, 0o600)
+	// if err != nil {
+	// 	return nil, fmt.Errorf("could not open device %s exclusively for writing", device)
+	// }
+	// return our disk
+	return initDisk(f, ReadWriteExclusive, opt.sectorSize)
+}
+
 // Create a Disk from a path to a device
 // Should pass a path to a block device e.g. /dev/sda or a path to a file /tmp/foo.img
 // The provided device must not exist at the time you call Create()
--- a/filesystem/fat32/fat32.go
+++ b/filesystem/fat32/fat32.go
@@ -3,6 +3,7 @@ package fat32
 import (
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"path"
 	"sort"
@@ -61,6 +62,10 @@ type FileSystem struct {
 	size            int64
 	start           int64
 	file            util.File
+}
+
+func (fs *FileSystem) Open(name string) (fs.File, error) {
+	return nil, fmt.Errorf("not implemented")
 }
 
 // Equal compare if two filesystems are equal
--- a/filesystem/fat32/file.go
+++ b/filesystem/fat32/file.go
@@ -3,6 +3,7 @@ package fat32
 import (
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 )
 
@@ -14,6 +15,14 @@ type File struct {
 	offset      int64
 	parent      *Directory
 	filesystem  *FileSystem
+}
+
+func (fl *File) ReadDir(n int) ([]fs.DirEntry, error) {
+	return nil, fmt.Errorf("not implemented")
+}
+
+func (fl *File) Stat() (fs.FileInfo, error) {
+	return nil, fmt.Errorf("not implemented")
 }
 
 // Read reads up to len(b) bytes from the File.
--- a/filesystem/file.go
+++ b/filesystem/file.go
@@ -1,11 +1,15 @@
 package filesystem
 
-import "io"
+import (
+	"io"
+	"io/fs"
+)
 
 // File a reference to a single file on disk
 type File interface {
-	io.ReadWriteSeeker
-	io.Closer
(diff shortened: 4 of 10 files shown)


deitch:
So each filesystem would implement `fs.FS`? 

> For me, there it is a dilemma: either to break the api and implement whole fs.FS support or implement wrapper functions. Wrappers look clumsily and, either way, require modifications of existing functions

I am not sure what you mean by this. Can you show a practical example?

aol-nnov:
> So each filesystem would implement fs.FS?

Yes, that's right! And please find the practical example in the related issue comment.

deitch:
I still don't get the wrapper issue?

aol-nnov:
Okay I'll elaborate more when I get to my laptop. I'm on the road now...

aol-nnov:
So, my thoughts on the implementation, as I see it. I may be wrong in some points as I'm still new to go.

As you can see from my wip patchset, there are very few receiver methods missing for the `diskfs.Filesystem` struct  to pass the `.(fs.FS)`  type assertion. The same stands for other fs-related structs.

1. `diskfs.File` was missing only `ReadDir()` and `Stat()` to look like `fs.File`
2. `diskfs.Filesystem` was missing `Open()` to look like `fs.FS`
3. and `iso9660.directoryEntry` was missing `Type()` and `Info()` to look like `os.FileInfo`, plus it has its receivers implemented as pointer receivers instead of value receivers. It did not work out at some point and I had to change them to value receivers.

Having such minor changes here I'm not tempted to write a wrapper, to be honest.

Now to incompatibilities: 
`diskfs.Filesystem` `OpenFile` receiver works almost as `fs.FS` `Open` should, but it should operate on directories, too. Plus, it must open an root directory of a filesystem when the path is `.` (as other fs implemetations do) and having two separate methons: for `Open` and `OpenFile` would be a 90% copy-paste of the latter.

Also, it is tempting to have [ReadDirFS](https://pkg.go.dev/io/fs#ReadDirFS) implemented, but `diskfs.Filesystem` `ReadDir` does not have proper return types for that..

So, here are my thoughts, pros and cons... Again, I might be totally wrong or missing something obvious, so please let me know!

deitch:
Now I understand where you are going.

I think you are in the right direction. My only issue is something that is not the fault of your implementation, but actually of go in general.

I cannot implement an interface by returning something that implements an interface, only the interface itself. Specifically, if `fs.FS` requires `Open(name string) (fs.File, error)`, then I must return actual `fs.File`, and not "something that implements `fs.File`". That means I cannot return my own thing. I cannot even return some other interface that includes this interface.

As an example, see [here](https://go.dev/play/p/4CoQNxYT8DP). Despite the fact that `impl` has `Hello() B`, which returns `B`, and `B` includes `A`, it will will not accept it.

So if I want to return `diskfs.File`, that will not work. I **must** return `fs.File`.

I am not sure I like that restriction. That may push us towards the idea of a wrapper. So you could have:

```go
filesystem.Open() filesystem.File
```

And maybe a wrapper, so you get (pseudo-code):

```go
fsys := diskfs.Open()
fsys.FS()   // convert to fs.FS
```

The reason I am concerned about having `Open()` just return an `fs.File`, is that it is [quite limited](https://pkg.go.dev/io/fs#File):

```go
type File interface {
	Stat() (FileInfo, error)
	Read([]byte), (int, error)
	Close() error
}
```

Whereas we have `ReadWriteSeeker` and `ReaderAt`. And there is no way to return `filesystem.File`, even if it fully implements `fs.File`.

AndreRenaud:
Given that generics have landed now, it might be possible to use the existing types internally, and then use a generic to convert them to fs.FS compatible types at compile time?

ie: https://play.golang.com/p/AHrNgLNYggU

This is a bit convoluted, but it does get around the issue that interfaces which define functions that return interfaces are not satisfied if an implementation returns a concrete type (that implements the correct interface).

AndreRenaud:
I had a play with the generics idea, and it didn't play out as nicely as I'd hoped. But using a wrapper type seems to work ok - would this approach be simpler? https://github.com/diskfs/go-diskfs/pull/184

deitch:
Yeah, I like what you did in #184 , let's move the detailed conversation over to there.

aol-nnov:
@AndreRenaud glad you came up with less invasive solution, thank you!

@deitch do you plan cutting a release any time soon? It's been almost a year since 1.3.0 and I'd be more than happy to see this feature included into the nearest release!

deitch:
Yeah, we can, I see no reason not to. 

Done

aol-nnov:
Thank you for the prompt response, @deitch !

PR #177, opened 2023-04-05
add rdrw open disk mode

ekodian:
Add RDWR mode for `diskfs.Open(...)`, so partitions could be resized without exclusive write access. For example **growpart** uses **sgdisk** to resize partitions on running systems.

### DIFF
--- a/diskfs.go
+++ b/diskfs.go
@@ -138,6 +138,8 @@ const (
 	ReadOnly OpenModeOption = iota
 	// ReadWriteExclusive open file in read-write exclusive mode
 	ReadWriteExclusive
+	// ReadWrite open file in read-write mode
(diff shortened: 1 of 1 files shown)


deitch:
Isn't that dangerous? When the kernel opens a disk, it almost always opens it exclusive, does it not?

ekodian:
Yes, but you **can** do it and for example **cloud-int** does this thing on start, using **growpart** to extend disk. I think - how you open disk is solely a user's responsibility.

deitch:
> solely a user's responsibility

Which is the argument I have made over and over again with other projects, and here I am arguing against it. Hoisted on my own petard. Good point!

deitch (approved):

PR #183, opened 2023-06-02
Use ASCIISet

elliotwutingfeng:
Proposing use of `ASCIISet` instead of `map[byte]bool`.

`ASCIISet` is a zero-dependency library for sets of ASCII characters with [28 times faster lookup speed](https://github.com/elliotwutingfeng/asciiset#results) than `map[byte]bool`.

### DIFF
--- a/filesystem/fat32/directoryentry.go
+++ b/filesystem/fat32/directoryentry.go
@@ -6,6 +6,8 @@ import (
 	"regexp"
 	"strings"
 	"time"
+
+	"github.com/elliotwutingfeng/asciiset"
 )
 
 // AccessRights is the byte mask representing access rights to a FAT file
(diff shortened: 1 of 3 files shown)


deitch:
I like this, but how can it be 28 times faster than a byte lookup in a map?

elliotwutingfeng:
Map uses a [hash function](https://cs.opensource.google/go/go/+/master:src/runtime/asm_amd64.s;l=1184?q=aeshashbody&ss=go%2Fgo) based on AES on most platforms, otherwise it falls back to [this](https://cs.opensource.google/go/go/+/master:src/runtime/hash64.go;l=25?q=memhashFallback&sq=&ss=go%2Fgo) hash function. This makes it compatible with more datatypes than just ASCII bytes.

Both the AES-like and fallback hash functions are fast, but not as fast as ASCIISet.Contains which uses just 2 bitwise and 3 arithmetic operations.

deitch (approved):

deitch:
Done. Thanks!

PR #184, opened 2023-06-03
Add fs.FS compatibility converter

AndreRenaud:
This is pretty rudimentary, but it appears to be enough to server the image via `http.FileServer`.

There is some poor naming in the code. If this approach appears to be basically acceptable, I'd welcome some feedback on cleaning it up a bit.

### DIFF
--- a/filesystem/compatibility.go
+++ b/filesystem/compatibility.go
@@ -1,34 +1,29 @@
-package converter
+package filesystem
 
 import (
 	"io/fs"
-	"net/http"
 	"os"
 	"path"
 	"time"
-
-	"github.com/diskfs/go-diskfs/filesystem"
 )
 
 type fsCompatible struct {
-	fs filesystem.FileSystem
+	fs FileSystem
 }
 
 type fsFileWrapper struct {
-	filesystem.File
+	File
 	stat os.FileInfo
 }
 
-type fsDirInfo struct {
-	name string
-}
+type fakeRootDir struct{}
 
-func (d *fsDirInfo) Name() string       { return d.name }
-func (d *fsDirInfo) Size() int64        { return 0 }
-func (d *fsDirInfo) Mode() fs.FileMode  { return 0 }
-func (d *fsDirInfo) ModTime() time.Time { return time.Now() }
-func (d *fsDirInfo) IsDir() bool        { return true }
-func (d *fsDirInfo) Sys() any           { return nil }
+func (d *fakeRootDir) Name() string       { return "/" }
+func (d *fakeRootDir) Size() int64        { return 0 }
+func (d *fakeRootDir) Mode() fs.FileMode  { return 0 }
+func (d *fakeRootDir) ModTime() time.Time { return time.Now() }
+func (d *fakeRootDir) IsDir() bool        { return true }
+func (d *fakeRootDir) Sys() any           { return nil }
 
 type fsDirWrapper struct {
 	name   string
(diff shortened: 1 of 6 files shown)


deitch:
This looks pretty good. 

I do wish we could have each implementation return FS, so we could get a clean-looking:

```go
fs, err := disk.CreateFilesystem(0, diskfs.TypeFat32)
fsys := fs.FS()
```

But the reality is that would be wasteful. It would require every filesystem implementation to implement `FS()`, when they all already have an interface they implement. So this approach you have here is superior:

```go
fs, err := disk.CreateFilesystem(0, diskfs.TypeFat32)
fsys := filesystem.FS(fs)
```

As for naming and structure, I see no need for this to be in its `package converter`. I would place it right at the root of `filesystem/`, so you can do what is above: `filesystem.FS(fs)`.

So maybe move this into [filesystem](https://github.com/diskfs/go-diskfs/tree/master/filesystem). The rest works fine.

It _does_ bother me that we cannot just have the returned filesystem from `CreateFIlesystem` be `fs.FS` compatible, but we cannot, as discussed [here](https://github.com/diskfs/go-diskfs/pull/175#issuecomment-1501146997), so let's stick with your wrapper.

Finally, I am confused by the need for [HTTPFS()](https://github.com/diskfs/go-diskfs/pull/184/files#diff-2713a6aa2f9ad63293d17c951ff60fc15eafa627e376995ae92e5e619d267412R126-R128). That seems extraneous to me. `http` package provides all we need, so just use it:

```go
fs, err := disk.CreateFilesystem(0, diskfs.TypeFat32)
fsys := filesystem.FS(fs)
httpfs := http.FS(fsys)
```

Adding the complexity by saving a single call doesn't buy us much.

AndreRenaud:
Another option might be to move this to an `internal` directory, and then add the `FS` method to each implementation, which calls this. So we don't add a new exposed function, just expand the `FileSystem` interface? If we put it in an `internal` directory however, that does mean that 3rd parties wouldn't be able to re-use it. Is that a concern?

This would mean you'd get something like:
```go
fs, err := disk.CreateFilesystem(0, diskfs.TypeFat32)
fsys := fs.FS()
```

But the fat32 implementation's function would just be (using the current naming - this will be changed soon):
```go
fat32.go:
func (fat32 *FilSsystem) FS() fs.ReadDirFS {
   return converter.FS(fat32)
}
```

The `HTTPFS()` function was just to avoid the extreme nesting in the following long line:
```go
http.Handle("/", http.FileServer(http.FS(converter.FS(fs))))
```
But I agree, it serves very little purpose, and the naming is horrible (`HTTPFS` is weirdly all caps). I'll drop it.

deitch:
This ready for review?

AndreRenaud:
Yes, I think so. The only question is whether it's worth adding the `func FS() fs.ReadDirFS` function to the `FileSystem` interface, and then implementing it using `filesystem.FS` for each of the 3 current implementations?

filesystem/compatibility.go
```diff
@@ -0,0 +1,116 @@
...
+	"time"
+)
+
+type fsCompatible struct {
+	fs FileSystem
+}
+
+type fsFileWrapper struct {
+	File
+	stat os.FileInfo
```
AndreRenaud, line 16: I've been using this with read-only images. But if this is to support read/write, then caching the `stat` info is probably a mistake.

deitch: It isn't. `fs.FS` currently is read-only. Even its few sub-variants are read-only. If it helps, I have implemented a full read-write FS variant (`fs.FS` compatible) which includes support for case-sensitivity and hardlinks and chmod and chown, even on underlying filesystems that do not support it. If you want a link, just ask.

deitch (approved):

deitch:
Your contribution is appreciated, @AndreRenaud !

filesystem/compatibility.go
AndreRenaud: Right, fs.FS is read-only, but `diskfs.FileSystem` can be read/write, so this abstraction could break?

deitch: I don't see how. `fs.FS` is just a subset of what is here. I am comfortable with what it is.

PR #186, opened 2023-09-14
iso9660: add basic symlink support

vanackere:
Hi, this commit adds basic symlink support for iso9660 filesystems with Rockridge extension.

(Far from perfect but allows to handle symlinks manually at least)

### DIFF
--- a/filesystem/iso9660/directoryentry.go
+++ b/filesystem/iso9660/directoryentry.go
@@ -482,7 +482,22 @@ func (de *directoryEntry) Size() int64 {
 
 // Mode() FileMode     // file mode bits
 func (de *directoryEntry) Mode() os.FileMode {
+	for _, ext := range de.extensions {
+		if s, ok := ext.(rockRidgeSymlink); ok && !s.continued {
+			return 0o755 | os.ModeSymlink
+		}
+	}
 	return 0o755
+}
+
+// Readlink tries to return the target link, only valid for symlinks
+func (de *directoryEntry) ReadLink() (string, bool) {
+	for _, ext := range de.extensions {
+		if s, ok := ext.(rockRidgeSymlink); ok && !s.continued {
+			return s.name, true
+		}
+	}
+	return "", false
 }
 
 // ModTime() time.Time // modification time


deitch:
Is that consumed anywhere? I do not see it.

The challenge is that the `directoryEntry` (not just for iso9660, but for all filesystems), implement [fs.FileInfo](https://pkg.go.dev/io/fs#FileInfo). There is no issue adding _more_ member functions to it; but it would need to be used somewhere to be useful. I was hoping the in-progress ext4 would show us something, but I cannot find the links in there.

I guess it would be ok, and since it is on the `iso9660.directoryEntry` and not a generic one, it works.

The only issue then, is, is it consumed anywhere?

vanackere:
Hello,

There is no public usage of this function since it is used only in an
internal utility I wrote to scan the content of debian ISOs and my code was
not working on this image :

https://cdimage.debian.org/cdimage/archive/10.13.0/amd64/iso-cd/debian-10.13.0-amd64-netinst.iso

Unfortunately the fs.FileInfo interface has no way to expose the relevant
information, therefore usage has to be done this way but for now this is
useful enough:

-------------------------------------
type ReadLink interface {
 ReadLink() (string, bool)
}

isSymlink := finfo.Mode()&fs.ModeSymlink != 0

if isSymlink {
if finfoRL, ok := finfo.(ReadLink); ok {
  link, ok := finfoRL.ReadLink()
  if ok {
   e.Link = link
  }
}
---------------------------

By the way, if you're refering to
https://github.com/trustelem/go-diskfs/commits/ext4-wip regarding to the
in-progress ext4: the read support is pretty good (including checksum
verifications... which was not obvious to debug) but I have no time for the
foreseeable future to work on the write part, nevertheway it may be useful
as it so just tell me if you're interested by a pull request !

Best regards,

Vincent

Le lun. 18 sept. 2023, 16:53, Avi Deitcher ***@***.***> a
écrit :

deitch:
Good point. OK, let's let CI run and see what it says.

deitch (approved):

PR #188, opened 2023-11-06
Open disk in O_SYNC mode when writing

DavidVentura:
I've had some intermittent issues with some NVMe disks ignoring the partitions I was writing with `diskfs.Partition` and immediately reading them back.
No amount of `syscalls.Sync` would make the issues go away, but changing the writing mode to `O_SYNC` did help.

### DIFF
--- a/diskfs.go
+++ b/diskfs.go
@@ -158,8 +158,8 @@ func (m OpenModeOption) String() string {
 
 var openModeOptions = map[OpenModeOption]int{
 	ReadOnly:           os.O_RDONLY,
-	ReadWriteExclusive: os.O_RDWR | os.O_EXCL,
-	ReadWrite:          os.O_RDWR,
+	ReadWriteExclusive: os.O_RDWR | os.O_EXCL | os.O_SYNC,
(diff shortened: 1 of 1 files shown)


deitch:
Wouldn't that make it terribly slow? It would force all writes not to return until the OS finished flushing to disk, as opposed to the usual disk behaviour.

If it is to address just the partition case, would a flush right after [this Write()](https://github.com/diskfs/go-diskfs/blob/master/disk/disk.go#L74) do it? And does the failure happen in that same function, say [here](https://github.com/diskfs/go-diskfs/blob/master/disk/disk.go#L82)?

DavidVentura:
The behavior I am seeing is that a `syscall.Sync()` call after `diskfs.File.Close()` is not always enough -- _something_ is not getting flushed. Even when calling `ReReadPartitions`, we do not see the new partition table take effect.

For me, it is indeed only for the partitioning use-case, which ends up writing very little so this sync behavior is not noticeable.

I'll try to reproduce with a `Flush` after writing the partition and let you know

PR #193, opened 2023-12-12
squashfs: fix off by one in directory counting

ncw:
This was causing "corrupted directory, had 257 entries instead of max
256" errors, however this was because parseDirectoryHeader had already
added one to the count.


deitch (approved):
Looks good, but won't pass CI until rebased because of the lint issues that were fixed a few hours ago.

ncw:
Rebased!

PR #194, opened 2023-12-12
squashfs: sometimes missing xattrs seems to be marked with 0xffffffffffffffff

ncw:
Using mksquashfs 1:4.5-3build1 and these flags causes the created
squash file not have xattrs but this is indicated by the index being
all 1s rather than the flag in the superblock.

mksquashfs $IMG $IMG.sqfs -comp zstd -Xcompression-level 3 -b 1M -no-xattrs -all-root


deitch (approved):
Looks good, thanks, but won't pass CI until rebased to catch the lint fix.

ncw:
Rebased!

PR #198, opened 2023-12-16
MBR: add partition table UUID and partition UUID

thirdeyenick:
This allows to retrieve the partition table UUID and partition UUID for MBR formatted disks. The partition table UUID (aka "disk identifier" or "disk signature") can be used to identify a disk when using the MBR format scheme. The partition UUID is then just the partition table UUID with the partitions index number appended.

### DIFF
--- a/partition/gpt/partition.go
+++ b/partition/gpt/partition.go
@@ -291,3 +291,8 @@ func (p *Partition) sectorSizes() (physical, logical int) {
 func (p *Partition) Equal(o *Partition) bool {
 	return p != nil && o != nil && *p == *o
 }
+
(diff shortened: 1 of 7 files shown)


deitch (approved):

deitch:
I was surprised that `mbr.initTable()` was unused, but apparently it was.

The UUID is optional in MBR, but it cannot hurt, so might as well.

Thank you for this.

deitch:
Some linting errors.

thirdeyenick:
> Some linting errors.

Should be fixed now.

deitch (approved):

PR #202, opened 2023-12-21
squashfs: cache the last block decompressed in the file for 10x speedup

ncw:
Before this change, we would spend an awful lot of CPU time reading
and decompressing the same block to only read a small part from it.

This change caches the last block used in the file handle and for
typical block sizes used in reads this makes a 10x difference in
throughput.


deitch (approved):

deitch:
This is great. Solid improvement.

PR #204, opened 2023-12-23
squashfs: factor read only OpenFile into Open method on directory entry

ncw:
This enables us to open the file directly from the directory listing which saves doing another directory traversal to find it again.

This uses a simple `Open` method modeled on the one in the zip archive [File.Open](https://pkg.go.dev/archive/zip#File.Open)

When extracting all the entries from the archive this makes a significant performance improvement as re-reading all the directory entries from the root is quite expensive.

### DIFF
--- a/filesystem/squashfs/directoryentry.go
+++ b/filesystem/squashfs/directoryentry.go
@@ -5,6 +5,8 @@ import (
 	"io/fs"
 	"os"
 	"time"
+
+	"github.com/diskfs/go-diskfs/filesystem"
 )
 
 // FileStat is the extended data underlying a single file, similar to https://golang.org/pkg/syscall/#Stat_t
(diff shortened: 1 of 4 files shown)


filesystem/squashfs/directoryentry.go
```diff
@@ -158,3 +161,39 @@ func (d *directoryEntry) Readlink() (string, error) {
...
+
+// Open returns an filesystem.File from which you can read the
+// contents of a file.
+//
+// Calling this on anything but a file will return an error.
+//
+// Calling this Open method is more efficient than calling
+// FileSystem.OpenFile as it doesn't have to find the file by
+// traversing the directory entries first.
+func (d *directoryEntry) Open() (filesystem.File, error) {
```
deitch, line 173: I was going to ask, why not have the `mode` arg here... and then realized, "we are talking about a read-only filesystem, so no point to it." Looks good.

deitch (approved):

PR #206, opened 2023-12-31
squashfs: add configurable block cache

### LINKED ISSUE #205: squashfs: caching of metadata and fragment blocks
I've been using this [3GB squashfs](https://drive.google.com/file/d/1Ho01W5H_UYibK599qXMKRUOU1y51Mecp/view?usp=drive_link) as a performance test.

With the `Open` optimization in #204 the timings are as follows to unpack the archive

- `unsquashfs` - 9s
- `rclone` with this library - 137s

That is quite a difference!

However I found putting a small LRU cache of metadata blocks and fragment blocks I can reduce that to

- `rclone` with this library and a bit of caching - 17s

Which is a huge difference. The problem is that it is quite CPU expensive to decompress the pages and that comes to dominate the runtime without caching.

I've done a proof of concept in the [fix-caching branch](https://github.com/ncw/go-diskfs/commits/fix-caching/).

This works extremely well so I think it is worth considering it for inclusion. However there are some open questions I'd like to discuss before submitting a PR.

## How should the cache be configured?

I've arbitrarily set the caches to hold to 64 metadata blocks, 64 fragments blocks cached for 10s. This can use a reasonable amount of memory(2`*`64`*`128k = 16M) but that seems affordable.

Should this be configured when creating the FileSystem? Or maybe the user should supply their own caching mechanism. An interface like

```go
type Cache interface {
    Get(pos int64) []byte
    Put(pos int64, buf []byte)
}
```

Would be enough. Once cache for metadata and one for fragments seems sensible. A cache for the data blocks isn't needed after caching the block in the file handle, but if you were re-opening files then it might be useful too.

## Memory use

In practice this seems to use a lot more memory than expected - I suspect it is the go garbage collector being slow to recycle the 128k blocks!

We should probably be recycling the memory blocks - a sync.Pool of `fs.blocksize` blocks seems sensible and that would probably help with the memory use. I'm not a huge fan sync.Pool having had lots of problems with it in rclone, but if it doesn't work there are lots of alternatives.

Doing this would involve a bit of re-plumbing but I think it would generate quite a small patch.

After the caching, memory copying is the top thing left in the profile!

## Caching library

My proof of concept uses the `github.com/hashicorp/golang-lru/v2/expirable` LRU cache which is widely used but not necessarily the right one.

I chose this just because it is already a dependency of rclone! However I'd be happy to use a different one or write a specific one.

## Anti-dogpile

If you look at the code you'll see a small anti-dogpile code (see [dogpile effect](http://highscalability.com/blog/2014/7/30/preventing-the-dogpile-effect-problem-and-solution.html)). This is absolutely essential when using the library multithreaded (like rclone does). There might be a library to do this, but the amount of code is small.

---- 
Anyway your thoughts much appreciated.

ncw:
This adds a configurable block cache as discussed in #205

This is done in 3 commits for ease of review - it compiles and the tests pass at each commit so they can be merged as 3 commits if desired.

With this cache this now performs within a factor of 2 of unsquashfs

- squashfs: make readMetadata a method on FileSystem
- squashfs: add LRU cache for blocks
- squashfs: use the LRU cache for fragments and metadata

### DIFF
--- a/filesystem/squashfs/lru.go
+++ b/filesystem/squashfs/lru.go
@@ -1,0 +1,138 @@
+package squashfs
+
+import (
+	"sync"
+)
+
+// A simple least recently used cache
+type lru struct {
+	mu        sync.Mutex
+	cache     map[int64]*lruBlock // cache keyed on block position in file
+	maxBlocks int                 // max number of blocks in cache
+	root      lruBlock            // root block in LRU circular list
+}
+
+// A data block to store in the lru cache
+type lruBlock struct {
+	mu   sync.Mutex // lock while fetching
+	data []byte     // data block - nil while being fetched
+	prev *lruBlock  // prev block in LRU list
+	next *lruBlock  // next block in LRU list
+	pos  int64      // position it was read off disk
+	size uint16     // compressed size on disk
+}
+
+// Create a new LRU cache of a maximum of maxBlocks blocks of size
+func newLRU(maxBlocks int) *lru {
+	l := &lru{
+		cache:     make(map[int64]*lruBlock),
+		maxBlocks: maxBlocks,
+		root: lruBlock{
+			pos: -1,
+		},
+	}
+	l.root.prev = &l.root // circularly link the root node
+	l.root.next = &l.root
+	return l
+}
+
+// Unlink the block from the list
+func (l *lru) unlink(block *lruBlock) {
+	block.prev.next = block.next
+	block.next.prev = block.prev
+	block.prev = nil
+	block.next = nil
+}
+
+// Pop a block from the end of the list
+func (l *lru) pop() *lruBlock {
+	block := l.root.prev
+	if block == &l.root {
+		panic("internal error: list empty")
+	}
+	l.unlink(block)
+	return block
+}
+
+// Add a block to the start of the list
+func (l *lru) push(block *lruBlock) {
+	oldHead := l.root.next
+	l.root.next = block
+	block.prev = &l.root
+	block.next = oldHead
+	oldHead.prev = block
+}
+
+// ensure there are no more than n blocks in the cache
+func (l *lru) trim(maxBlocks int) {
+	for len(l.cache) > maxBlocks && len(l.cache) > 0 {
+		// Remove a block from the cache
+		block := l.pop()
+		delete(l.cache, block.pos)
+	}
+}
+
+// add block to the cache, pruning the cache as appropriate
+func (l *lru) add(block *lruBlock) {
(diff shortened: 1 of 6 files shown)


deitch (approved):
I wouldn't mind some performance test on this, but not for me to complain.

Do you want to add a comment somewhere (likely in squashfs.go, or in doc.go) about why we have a cache for this, capture some of your thoughts from the PR and issue?

ncw:
> I wouldn't mind some performance test on this, but not for me to complain.

Were you thinking of a Go benchmark?

Here are some with rclone - unpacking my favourite tensorflow 3GB squashfs with rclone

| no caching | 256 MB cache | 128MB cache | 64 MB cache | 32 MB cache |
| ---- | ---- | ---- | ---- | ---- |
| 206s | 16.7s | 17.5s | 23.4s | 54.s |

Whereas `unsquashfs` takes `12.0s`
 
> Do you want to add a comment somewhere (likely in squashfs.go, or in doc.go) about why we have a cache for this, capture some of your thoughts from the PR and issue?

I've added some notes to the Filesystem.Read method which is where the cache gets initialized - this will appear in the Go doc and be hopefully obvious to the users.

Interestingly the [squashfs kernel module](https://docs.kernel.org/filesystems/squashfs.html#squashfs-internal-cache) uses the same design choice (which I hadn't noticed before) about having a fragment and metadata cache.

Note that this PR will conflict with #201 I think, which I'll fix up.

ncw:
> Note that this PR will conflict with https://github.com/diskfs/go-diskfs/pull/201 I think, which I'll fix up.

Oooh, you merged that while I was writing the above! It is now fixed.

deitch:
I think this is solid.

Do you want to throw a benchmarking doc in there? You captured much of it here, but this tends to get lost. Specifically either in the `doc.go`, or in the `lru.go`? Just showing how you created the file, how you used rclone, so we have some basis for understanding it in the future?

ncw:
I added some info about benchmarks here

https://github.com/ncw/go-diskfs/blob/a7c64791e8ccc64f3bed0e0618e712224a1338b5/filesystem/squashfs/squashfs.go#L126-L141

What do you think - look ok?

ncw:
Actually I forgot to say how I made the file, let me add that!

ncw:
Done!

https://github.com/ncw/go-diskfs/blob/84de30b32142fa0fec9e29ab378e0a83076a3fc3/filesystem/squashfs/squashfs.go#L126-L147

deitch (approved):

PR #216, opened 2024-05-22
iso9660 finalizeFileInfo contain enough info for rockridge extensions

### LINKED ISSUE #215: Finalize error with a combination of RockRidge and ElTorito
The following code fails with:
`
unable to calculate size of directory for .: could not calculate child boot.catalog entry size .: could not convert to dirEntry: error getting extensions for IEEE_P1282 at path boot.catalog: error reading file /var/folders/lb/nhsppbxs5_v27pn_x3kf66wc0000gn/T/diskfs_iso1583485680/boot.catalog: lstat /var/folders/lb/nhsppbxs5_v27pn_x3kf66wc0000gn/T/diskfs_iso1583485680/boot.catalog: no such file or directory
`

```
fs, _ := iso9660.Create(f, 0, 0, blocksize, "")
file, _ := fs.OpenFile("/limine-bios-cd.bin", os.O_CREATE|os.O_RDWR)
defer file.Close()
_ = fs.Finalize(iso9660.FinalizeOptions{
	RockRidge: true,
	ElTorito: &iso9660.ElTorito{
		Entries: []*iso9660.ElToritoEntry{
			{
				Platform:  iso9660.BIOS,
				Emulation: iso9660.NoEmulation,
				BootFile:  "/limine-bs-cd.bin",
				BootTable: true,
				LoadSize:  4,
			},
		},
	},
})
```

deitch:
The rock ridge extensions at various places assumed all files were actual files in the workspace. However, `finalizeFileInfo` supports in-memory files.

The solution is to extend `finalizeFileInfo` to have the additional features rock ridge is looking for: accessTime, modTime, linkTarget. `finalizeFileInfo` already implements `fs.FileInfo`. Then we can convert the call to RockRidge to just accept `finalizeFileInfo` and get the properties.

My tests show it works.

Fixes #215

### DIFF
--- a/filesystem/iso9660/directoryentrysystemuseextension.go
+++ b/filesystem/iso9660/directoryentrysystemuseextension.go
@@ -37,7 +37,7 @@ type suspExtension interface {
 	Descriptor() string
 	Source() string
 	Version() uint8
-	GetFileExtensions(string, bool, bool) ([]directoryEntrySystemUseExtension, error)
+	GetFileExtensions(*finalizeFileInfo, bool, bool) ([]directoryEntrySystemUseExtension, error)
 	GetFinalizeExtensions(*finalizeFileInfo) ([]directoryEntrySystemUseExtension, error)
 	Relocatable() bool
 	Relocate(map[string]*finalizeFileInfo) ([]*finalizeFileInfo, map[string]*finalizeFileInfo, error)
--- a/filesystem/iso9660/finalize.go
+++ b/filesystem/iso9660/finalize.go
@@ -3,6 +3,7 @@ package iso9660
 import (
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path"
 	"path/filepath"
@@ -12,11 +13,13 @@ import (
 	"time"
 
 	"github.com/diskfs/go-diskfs/util"
+	"github.com/djherbis/times"
 )
 
 const (
 	dataStartSector         = 16
 	defaultVolumeIdentifier = "ISOIMAGE"
+	elToritoBootTableOffset = 8
 )
 
 // FinalizeOptions options to pass to finalize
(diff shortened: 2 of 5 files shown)


pierrec:
In my app, I get the following error:
```
2024/05/22 17:13:53 error copying file boot/limine-bios-cd.bin to disk, copied 20480 bytes, expected 20536
panic: error copying file boot/limine-bios-cd.bin to disk, copied 20480 bytes, expected 20536
```

Which is triggered by [this check](https://github.com/diskfs/go-diskfs/blob/rock-ridge-catalog-fix/filesystem/iso9660/finalize.go#L734).
If I disable it, then it completes and the produced .iso boots up!

pierrec:
Note the invalid value for the expected one in the error message.

deitch:
> In my app, I get the following error:
> Which is triggered by [this check](https://github.com/diskfs/go-diskfs/blob/rock-ridge-catalog-fix/filesystem/iso9660/finalize.go#L734).

That check should make sense. With every file, we expect it to copy either the file size, or file size plus boot table size, if we added it.

How can we replicate your issue?

pierrec:
Let me put itt up somewhere. Playing around with which files I included or changing the name of the one failing, does pass in some cases.

pierrec:
https://github.com/pierrec/go-diskfs-issue

deitch:
That helped.

Here is what I do not remember, as it has been years since I did the iso9660 implementation.

When a record has a boot table, i.e.:

```go
		ElTorito: &iso9660.ElTorito{
			BootCatalog: "boot.cat",
			Entries: []*iso9660.ElToritoEntry{
				{
					Platform:  iso9660.BIOS,
					Emulation: iso9660.NoEmulation,
					BootFile:  "/boot/limine-bios-cd.bin",
					BootTable: true,
					LoadSize:  4,
				},
				{
					Platform:  iso9660.EFI,
					Emulation: iso9660.NoEmulation,
					BootFile:  "/boot/limine-uefi-cd.bin",
					BootTable: true,
				},
			},
		},
```

`BootTable` is set to `true`.

In that case, it generates the boot table and inserts it into the file from offset 8 for 56 bytes.

Is that supposed to _override_ the data in the file at those bytes, keeping the file the same size? Or _insert_ it into the file at that location, thus extending the file by 56 bytes?

The code from years ago implies that it overwrites it, which is why your sample is failing, as the test later assumes that it is inserted.

So I am going to put back the original test, just allowing for the case that the original file is <64 bytes.

pierrec:
Success!
Very nice and your prompt support very much appreciated!!

pierrec (approved):

PR #220, opened 2024-05-29
Fix off-by-one error in GPT last data sector

rjosephwright:
The last usable data sector should be the sector of the start of the backup partition table - 1, but it does not subtract the 1. This fixes it and adds an additional test with sgdisk to verify.


deitch (approved):
I particularly like the test. Thank you.

PR #222, opened 2024-05-31
GPT table repair should reset last data sector

rjosephwright:
This modifies the GPT table Repair method to update the last data sector and the Verify method to check for this. If this is not done, it can result in e.g. sgdisk warning of a partition overlapping the last data sector when the disk is resized.

A test has been added to create a disk with a GPT table and verify it, then resize and repair it and verify again successfully.


deitch (approved):
Makes sense. Thank you for the addition.

PR #233, opened 2024-07-03
implement remove and rename feature for fat32

### LINKED ISSUE #191: Support to remove files
Hi, thanks for the awesome library.

I played around a bit with the library and was able to create new files and upate existing files on a fat32 partition inside an image. Unfortunately, there seems to be no support for removing files. Do I miss something? If its really missing, will the remove functionality come in the future?

Thanks and best regards

ludelion5:
Sorry for the late PR. Here is my proposal to rename and remove files in fat32.

closes #191


deitch (commented):
Overall, this looks great, and thank you.

In terms of naming, are these meant to be restricted just to _files_, or any structure (files, dirs, symlinks where supported, etc.)? I would think the latter, since CLI-level commands for the above in most OSes don't care. Same for linux `mv` / Windows `ren`.

So should these be `RemoveFile()` and `RenameFile()`, or just `Remove()` and `Rename()`?

On the other hand, in keeping with golang style, maybe these should parallel and be [`Remove()`](https://pkg.go.dev/os#Remove) and [`RemoveAll()`](https://pkg.go.dev/os#RemoveAll) and [`Rename()`](https://pkg.go.dev/os#Rename]?

ludelion5:
Thanks for the feedback. I will look into your request and test, if the implementation will also work for symlinks and directories. If so, its a good idea to stick to your suggested naming-scheme.

Additionally I can check if the RemoveAll is easy to implement, I have to see if I can spare enough time for that feature.

ludelion5:
Sorry for the late update on this PR.

We adapted our Rename and Remove methods to the new interface.

Additionally we have implemented the ability to remove empty directories and the renaming to already existing files (overwriting).
Currently, the renaming (move) works only in the same directory, this would require a deeper understanding of your `fs.allocateSpace`, I guess this method has to be called multiple times, depending on how many directory levels the file moves. We check if the base-dir is different and return an error.

filesystem/fat32/fat32.go
```diff
@@ -637,18 +637,130 @@ func (fs *FileSystem) OpenFile(p string, flag int) (filesystem.File, error) {
...
-// Rename renames (moves) oldpath to newpath. If newpath already exists and is not a directory, Rename replaces it.
-//
-//nolint:revive // parameters will be used eventually
-func (fs *FileSystem) Rename(oldpath, newpath string) error {
-	return filesystem.ErrNotImplemented
-}
-
 // removes the named file or (empty) directory.
 //
 //nolint:revive // parameters will be used eventually
```
aol-nnov, line 642: As the method got implemented, lint exception is not needed anymore here. Please remove it.

deitch:
> Currently, the renaming (move) works only in the same directory, this would require a deeper understanding of your fs.allocateSpace, I guess this method has to be called multiple times, depending on how many directory levels the file moves. We check if the base-dir is different and return an error

That certainly is enough for now. We go from "this is not implemented" to "this is partially implemented", which his an improvement. So thank you.

deitch (changes requested):
Just the lint comment.

ludelion5:
done

deitch (approved):

deitch:
Just some lint errors that need to be fixed.

ludelion5:
`make lint` should now be satisfied

deitch (approved):

deitch:
Looks great, thanks!

ludelion5:
are you planning to create a new tag as well?

deitch:
Yeah but let's wait for this one.

ludelion5:
Do you have an estimation when a new tag is created?

deitch:
https://github.com/diskfs/go-diskfs/releases/tag/v1.5.0

PR #234, opened 2024-07-08
Removed go.uuid dependency

danielh2942:
github.com/google/uuid was already a dependency so I removed the github.com/satori/go.uuid dependency that was introduced recently 

satori/go.uuid looks to be no longer maintained as of 2018


deitch (approved):
Thank you for this.

PR #235, opened 2024-07-10
Allow resizing a GPT

paul-rodriguez:
These changes allow:
* observing the total size and last data sector of a GPT
* resizing a GPT (for example to make it fit its block device)
* expanding a GPT partition

### DIFF
--- a/partition/gpt/partition.go
+++ b/partition/gpt/partition.go
@@ -296,3 +296,9 @@ func (p *Partition) Equal(o *Partition) bool {
 func (p *Partition) UUID() string {
 	return p.GUID
 }
+
+// Expand increases the size of the partition by a number of sectors
+func (p *Partition) Expand(sectors uint64) {
+	p.End += sectors
+	p.Size += sectors * uint64(p.logicalSectorSize)
+}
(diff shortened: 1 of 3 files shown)


paul-rodriguez:
> The assumption is that one would expand the partition, and then call `table.Write()` to write it to disk?

Yes indeed.

> Can we add some tests? Look at how `table_test.go` does it, which probably is the right place for it. I wouldn't want to affect the original `testdata/gpt.img`, but maybe it can be copied to a tmpdir, then modified, and read using the library to see that the changes took? Since we already test the reading of partition tables independently, we can rely on those to validate the changes.

Sure thing, thanks for the pointers.

paul-rodriguez:
@deitch I have pushed some unit tests for the Resize and Expand functions. My apologies if this mention is not good etiquette!

deitch (dismissed):

deitch:
Not a problem at all. Thank you for this. Let's let CI run.

deitch:
Some linting issues.

paul-rodriguez:
It should be ok now

deitch:
Looks good. Are you ready to go with this?

deitch (approved):

paul-rodriguez:
Yes, it works fine for my purposes. Thank you!

deitch:
Thanks @paul-rodriguez

PR #236, opened 2024-07-16
doc: remove outdated example in README, point to real source instead

quite:
The example inlined in the README seems to be for an older version of the package. Let's not try to maintain it, pointing at the example file instead.


quite:
@deitch here is a small fix to the README, to avoid confusing new users

deitch (approved):

deitch:
Heh, definitely better. Thank you.

Issue #237, opened 2024-07-18
Use of errors.Unwrap makes it harder to use the API


quite:
In https://github.com/diskfs/go-diskfs/pull/164 a couple of errors that the `Create` function can run into are unwrapped before being returned as part of an extended errors string. I'm sure that the intent was well-meaning. But I think that it makes things harder for me as a *user* of the API that the diskfs package provides.

I was about to use `os.IsExist()` on the error returned by `Create`, and couldn't understand why it would not be true...

I think that wrapping errors is the more sensible thing to do. Users of this API can then dig around as they wish for things to report to *user* of a program. But this is not the task of the API. Especially since `initDisk` is unexported so I can only use it through `Create`.

In the first instance, when `OpenFile` errors, I think `Create` should do something like:

```go
return nil, fmt.Errorf("OpenFile failed: %w", err)
```

A caller of `Create` may on error do: `return fmt.Errorf("Create failed: %w", err)`. And the printing of this would look like this (if the device was `bootable.iso`):

```
Create failed: OpenFile failed: open bootable.iso: file exists
```

This way the error does not drop any error info collected on the way. And API user can do as the wish. As it stands, I'll have to first check existence of the file myself before calling `Create` (well I'm not gonna do string compare on the error string :) Feels unnecessary really.

What do you think @shellwhale and @deitch

deitch:
Yeah, this makes sense. I think the original motivation there was, "we are dropping errors, let us add them into the response". This just brings it one step further; if we properly wrap them, we get the information _and_ the ability to work with it as an error. Open a PR?

PR #240, opened 2024-07-21
Wrap errors, not Unwrap

quite:
Fixes https://github.com/diskfs/go-diskfs/issues/237

### DIFF
--- a/diskfs.go
+++ b/diskfs.go
@@ -344,11 +344,11 @@ func Create(device string, size int64, _ Format, sectorSize SectorSize) (*disk.D
 	}
 	f, err := os.OpenFile(device, os.O_RDWR|os.O_EXCL|os.O_CREATE, 0o666)
 	if err != nil {
-		return nil, fmt.Errorf("could not create device %s: %v", device, errors.Unwrap(err))
+		return nil, fmt.Errorf("could not create device %s: %w", device, err)
 	}
 	err = os.Truncate(device, size)
 	if err != nil {
-		return nil, fmt.Errorf("could not expand device %s to size %d: %v", device, size, errors.Unwrap(err))
+		return nil, fmt.Errorf("could not expand device %s to size %d: %w", device, size, err)
 	}
 	// return our disk
 	return initDisk(f, ReadWriteExclusive, sectorSize)


deitch:
There are 2 tests failing, both related to converting superblock to bytes. Yet you didn't change anything there.

Can you run the tests on the `master` branch and on your own locally? You just need to check `go test ./filesystem/ext4`

quite:
> go test ./filesystem/ext4

After running the `./buildimg.sh` they pass on both

deitch:
Then why doesn't it pass in CI? 🤷‍♂️

quite:
Something is up with that runner/image? Same here now https://github.com/diskfs/go-diskfs/pull/241

deitch:
> After running the ./buildimg.sh they pass on both

Same here. You don't need to run it manually, anyways. Just remove the contents of (or the directory itself) `filesystem/ext4/testdata/dist/` and it will autogenerate it.

deitch:
I just ran a test on macOS and on Ubuntu 22.04. It shouldn't matter, since the actual images generating from within `buildimg.sh` are run inside docker, but I did it anyways. Everything passes.

deitch:
There are 2 sections of the superblock that are mismatching. The second is just the checksum, which is different because of the first. Resolve the first, the checksum issue should go away.

The first is total bytes written. The actual is showing 0x9400 = 37888, while the expected is 0x95da = 38362, a difference of 474 (the total KiB written is at position 0x178 in the superblock, little-endian).

Ah, I see it, the issue is [here](https://github.com/diskfs/go-diskfs/blob/ec697b09567e4458ea02a5e2e64cbc792fc35389/filesystem/ext4/common_test.go#L675-L680). After `buildimg.sh` builds the filesystem, we need some way to know the actual value there, so we can test that diskfs is reading it correctly. It uses the output of `dumpe2fs ext4.img > stats.txt`, which also is in `testdata/dist/`. The test utility `testGetValidSuperblockAndGDTs()` reads the output of that. For example, the line from the `stats.txt` file that gives total written:

```
Lifetime writes:          37 MB
```

Unfortunately, that is in MB and not KB, and is not precise. So a small variant can throw it off. We try to catch it [here](https://github.com/diskfs/go-diskfs/blob/ec697b09567e4458ea02a5e2e64cbc792fc35389/filesystem/ext4/common_test.go#L675-L680):

```go
	sbKBWritten := float64(sb.totalKBWritten)
	parsedKBWritten := float64(parsed.totalKBWritten)
	KBdiff := math.Abs(parsedKBWritten - sbKBWritten)
	if KBdiff/sbKBWritten < 0.01 {
		sb.totalKBWritten = parsed.totalKBWritten
	}
```

But it is a bit of a guessing game. It would be good if `dumpe2fs` (or some other reliable tool) gave precise stats in KB written. `debugfs -R` gives the same output.

deitch:
#242 fixes this, it has been merged in. Rebase on master and try again.

deitch (approved):

PR #243, opened 2024-07-26
writable ext4

deitch:
This is a first draft of write functionality for ext4, including tests (of course). The failure cases all work as expected. There do not appear to be any missing write-implementing functions, or any such functions that have missing functionality. However, some of that logic almost certainly is incorrect, as proven by the testing errors.

Anyone who want to assist fixing the functions is welcome.


deitch:
This works, surprisingly well. I suspect it will break down with:

* splitting/extending the extents tree from the inode
* creating a brand-new ext4 filesystem

But I am happy to merge it in as is, and fix individual issues as they arise.

PR #251, opened 2024-08-31
Closes #250: fat32 8.3 lowercase names, make fs.OpenFile case insensitive for 8.3 and long filenames

### LINKED ISSUE #250: fat32 reads/writes incorrect 8.3 lowercase flags, should be case insensitive on fs.OpenFile
There are two distinct but related issues:

1. Lowercase filenames created with mtools are read as "lower.UPP" by go-diskfs, ie "syslinux.cfg" is read as "syslinux.CFG".  This is kind of a cosmetic issue, except for issue 2.
2. Because os.OpenFile doesn't ignore case when opening a file, the filename returned from above cannot be found (ie, attempt to open 'syslinux.CFG' as that was returned by fs.ReadDir, but get file not found instead)

Related to 2, I also added a test for mixed case on long file names in fs.OpenFile (verified on Linux vfat with "touch some-long-file && cat SOME-LONG-FILE").

I have a PR to raise for these issues, with unit tests.

jsando:
Let me know if any changes are required.

I wasn't sure if adding another 10MB fat32 image built with mtools was the best way to go for the unit tests but was a little confused when I tried to add it to the existing fat32.img file, it seems that file is not built by the shell script in the same folder.

### DIFF
--- a/filesystem/fat32/directoryentry.go
+++ b/filesystem/fat32/directoryentry.go
@@ -103,7 +103,7 @@ func (de *directoryEntry) toBytes() ([]byte, error) {
 	}
 
 	if de.lowercaseExtension {
-		dosBytes[12] |= 0x04
+		dosBytes[12] |= 0x10
 	}
 	if de.lowercaseShortname {
 		dosBytes[12] |= 0x08
@@ -161,7 +161,7 @@ byteLoop:
 		isArchiveDirty := b[i+11]&0x20 == 0x20
 		isVolumeLabel := b[i+11]&0x08 == 0x08
 		lowercaseShortname := b[i+12]&0x08 == 0x08
-		lowercaseExtension := b[i+12]&0x04 == 0x04
+		lowercaseExtension := b[i+12]&0x10 == 0x10
 
 		entry := directoryEntry{
 			filenameLong:       lfn,
--- a/filesystem/fat32/fat32.go
+++ b/filesystem/fat32/fat32.go
@@ -511,7 +511,7 @@ func (fs *FileSystem) ReadDir(p string) ([]os.FileInfo, error) {
 		}
 		fileExtension := e.fileExtension
 		if e.lowercaseExtension {
-			shortName = strings.ToLower(fileExtension)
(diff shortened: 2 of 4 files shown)


deitch:
> I wasn't sure if adding another 10MB fat32 image built with mtools was the best way to go for the unit tests

You are right, it isn't, we should just expand the existing image with some more entries to test these.

> but was a little confused when I tried to add it to the existing fat32.img file, it seems that file is not built by the shell script in the same folder.

Yeah. The script to generate it is inline in the `testdata/README.md`. I had meant to have the script generate it. The problem is that the tests then fail. It is highly sensitive to it, and each regeneration ends up requiring modifying the tests. If you look at ext4, in contrast, it uses standard CLI tools to extract information about the filesystem, and the tests consume those as the expected value. It was not as easy to do with fat32. Maybe I will try, see if I can get it to work.

deitch:
Yeah, I am getting somewhere. Hang tight.

deitch:
Done. @jsando rebase on latest, please. The entire `fat32.img` now is auto-generated with tests, very similar to what I did with ext4. It uses standard tools to inspect it, and then the tests read those outputs to compare with its own values. Works quite well.

You should be able to modify `mkfat32.sh` as you need.

jsando:
Rebased and pushed.   That was a nice change you did to delete the fat32 test image from git and generate it as part of the build.

Also I saw your comment on the ticket, I will add more information.

deitch:
> That was a nice change you did to delete the fat32 test image from git and generate it as part of the build.

Yeah, it makes things much easier. When initially creating this, you go for easy, or not _too_ much additional effort. At a certain point, the effort of changing all of your fixed test structures because you regenerated one test file becomes too much. Too most of a day, but done.

I like this. The tests are easy to read and understand.

deitch (approved):

deitch:
Let CI run, and when clean, we can let it in.

deitch:
And in. Thank you for finding and fixing this.

deitch:
I am most curious to know where you are using this. I will drop an email line.

PR #255, opened 2024-09-10
Speed up fat32 impl

Frostman:
Hello,

We're looking to use go-diskfs to build an air-gapped installer image for our product. It includes a bunch of files (like grub tree), OCI images, binaries and etc. 

The speed up is 20+ times on bigger file systems - from 540 seconds to 23 seconds for a 1G image with two partitions of ~500MB each.

The test setup is:

```
Intel(R) Core(TM) Ultra 7 155H, 32GiB RAM, SN740 NVMe WD 512GB
go 1.23.0
6.10.7-200.fc40.x86_64 #1 SMP PREEMPT_DYNAMIC
```

Here are three profiling reports:

- Unmodified: [v1.4.1.pdf](https://github.com/user-attachments/files/16949444/v1.4.1.pdf)
- Removing keys sort: [v1.4.1-nosort.pdf](https://github.com/user-attachments/files/16949443/v1.4.1-nosort.pdf)
- Replacing map with slice: [v1.4.1-nomap.pdf](https://github.com/user-attachments/files/16949442/v1.4.1-nomap.pdf)

### DIFF
--- a/filesystem/fat32/common_test.go
+++ b/filesystem/fat32/common_test.go
@@ -467,7 +467,7 @@ func testReadFilesystemData() (info *testFSInfo, err error) {
 				rootDirCluster: 2,         // root is at cluster 2
 				size:           sizeInBytes,
 				maxCluster:     numClusters,
-				clusters:       make(map[uint32]uint32),
+				clusters:       make([]uint32, numClusters+1),
 			}
 		case inClusters && len(clusterLineMatch) > 4:
 			start, err := strconv.Atoi(clusterLineMatch[1])
--- a/filesystem/fat32/fat32.go
+++ b/filesystem/fat32/fat32.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"os"
 	"path"
-	"sort"
 	"strings"
 	"time"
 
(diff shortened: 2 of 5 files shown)


deitch:
20x! That is great. I will kick off CI here, and take a closer look tomorrow.

deitch (approved):

Frostman:
@deitch, thx for merging!

Removing unused keys sort gave "only" 3x speed up, and that's b/c sort is not too slow, I think building list of keys() before sort was the slower part. If you'd look at the profile, you can see that `mapaccess2_fast32` called from `allocateSpace` is the biggest time consumer and that is just implementation of the get from the map, source https://go.dev/src/runtime/map_fast32.go. The problem is that read from the map is significantly more expensive then read from slice by index. When you're reading from map you need to calculate hash and go through hash collisions/overflow. When you're reading from slice it's simply memory pointer offset.

Frostman:
@deitch the project I've mentioned is an installer for the https://hedgehog.cloud/. We're a startup building a product for DCs.

deitch:
> project I've mentioned is an installer for the https://hedgehog.cloud/. We're a startup building a product for DCs

That is pretty cool. I will reach out offline to ask about it.

PR #257, opened 2024-09-19
bump go version, add linter fixes

deitch:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/diskfs/go-diskfs
 
-go 1.21
+go 1.23
```
AkihiroSuda, line 3: Was there any reason to drop the support for Go 1.22?

deitch: Not particularly. It kind of went along with other flows. We can step it back. Open a PR and make sure it passes lint and tests, and happy to merge it in.

AkihiroSuda: - https://github.com/diskfs/go-diskfs/pull/262

PR #260, opened 2024-09-26
fix offset when reading and writing fat32 files to avoid overruns

deitch:
Fixes the issue raised by @mrbojangles3. 

With thanks to @mrbojangles3, whose tests are now incorporated into `fat32_test.go`

### DIFF
--- a/filesystem/fat32/fat32_test.go
+++ b/filesystem/fat32/fat32_test.go
@@ -7,13 +7,17 @@ package fat32_test
 
 import (
 	"bytes"
-	"crypto/rand"
 	"fmt"
 	"io"
+	"math/rand/v2"
 	"os"
+	"path"
+	"path/filepath"
 	"strings"
 	"testing"
 
+	"github.com/diskfs/go-diskfs"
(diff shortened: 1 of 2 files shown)


filesystem/fat32/fat32_test.go
```diff
@@ -258,7 +264,7 @@ func TestFat32Read(t *testing.T) {
 				corrupted := ""
 				if tt.bytechange >= 0 {
 					b := make([]byte, 1)
-					_, _ = rand.Read(b)
+					_, _ = chacha.Read(b)
```
AkihiroSuda, line 267: Any reason to switch away from `crypto/rand` to `math/rand/v2` ?
https://pkg.go.dev/math/rand/v2#ChaCha8.Read is not available for Go 1.22
- https://github.com/diskfs/go-diskfs/pull/262

deitch: Nope. It was done because lint was complaining when go1.23. With 1.22 we should not switch.

PR #263, opened 2024-10-08
fat32: getter for file content offset

malt3:
This allows getting the raw offset to a contiguous file. Helpful to construct an `io.Reader` or `io.ReadSeeker`.

### DIFF
--- a/filesystem/fat32/file.go
+++ b/filesystem/fat32/file.go
@@ -1,6 +1,7 @@
 package fat32
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -206,3 +207,35 @@ func (fl *File) Close() error {
 	fl.filesystem = nil
 	return nil
 }
+
+func (fl *File) GetContentSection() (int64, int64, error) {
+	fs := fl.filesystem
+	bytesPerCluster := fs.bytesPerCluster
+	clusters, err := fs.getClusterList(fl.clusterLocation)
+	start := int(fs.dataStart)
+	if fl.fileSize == 0 {
+		return 0, 0, nil
+	}
+	if err != nil {
+		return 0, 0, fmt.Errorf("unable to get list of clusters for file: %v", err)
+	}
+	if len(clusters) == 0 {
+		return 0, 0, fmt.Errorf("file has no clusters")
+	}
+	offset := uint32(start) + (clusters[0]-2)*uint32(bytesPerCluster)
+	size := int64(bytesPerCluster)
+
+	// check if clusters are contiguous
+	last := clusters[0]
+	for _, cluster := range clusters[1:] {
+		if cluster != last+1 {
+			return int64(offset), size, ErrNotContiguous
+		}
+		last = cluster
+		size += int64(bytesPerCluster)
+	}
+
+	return int64(offset), int64(fl.fileSize), nil
+}
+
+var ErrNotContiguous = errors.New("file is not contiguous")


deitch:
Thanks for the submission @malt3 

I do not quite get what `GetContentSection()` does? It returns 2x int64 and an error? What do those represent?

malt3:
It returns the absolute offset and the file size on success. 
Those can be used in the constructor of a `io.SectionReader`: https://pkg.go.dev/io#NewSectionReader

For the range to be valid, I test that the file is contiguous (all blocks are next to each other in the correct order). If the file cannot be read as a contiguous range, the function returns an error.

deitch:
> absolute offset and file size

I still am having a hard time understanding. `io.SectionReader` is passed an `io.ReaderAt`, offset within the ReaderAt and max bytes to read. If working against, say, your local ext4 or APFS or whatever, you would do `os.Open()` to get an `os.File`, and then pass it to `NewSectionReader(f, offset, max)`. In this case, AFAIUI, `max` is not supposed to be the max size of the file (`ReaderAt` would return `io.EOF` when you get there anyways), just the max number of bytes to return, based on your desires, and the offset.

```go
f, err := os.Open("/path/to/file")
r := io.NewSectionReader(f, 25, 6000) // or whatever numbers you want
```

Why wouldn't you do the same here?

```go
f, err := fs.OpenFile("/path/to/file", flag)
r := io.NewSectionReader(f, 25, 6000) // or whatever numbers you want
```

Why is `GetContentSection()` needed, rather than just passing the `File` to `io.NewSectionReader()`. You don't need to worry about contiguous or anything like that. The `File` structure hides it all for you, giving you a simple `Read()` abstraction.

malt3:
I do not want to use the `Read` abstraction built into the `File` struct. I want to construct my own `ReaderAt` (and possibly `WriterAt`) for a file from the underlying block device or file that contains the filesystem. 
If I now that the file is contiguous, it is safe to do so.

deitch:
Why would you? The whole point of the abstraction is so that you don't have to deal with things like figuring out clusters. If you opened a real fat32 file system (or ext4, etc), you wouldn't do that. 

What's the benefit here?

malt3:
The real benefit is that this allows me to modify file contents in place.

deitch:
> modify file contents in place

what does that mean? What is the process you are doing, and what is the parallel to it with a normal mounted filesystem?

malt3:
This api is very similar to [fiemaps](https://www.kernel.org/doc/Documentation/filesystems/fiemap.txt) in linux. It gives me direct access to extents of a file, without changing the structure of the filesystem.

I the end, a filesystem is just a bag of bytes, and the file contents are stored somewhere inside. I want to be able to change the contents of a file without modifying any metadata or layout.
The above mentioned API allows to do this, but is more complete: Instead of enforcing that the whole file is contiguous, it allows the user to see the offsets of individual extends of a file.

deitch:
> Similarly, mmap allows user space to get random read / write access into a file

But doesn't mmap work through the filesystem driver? You get a file descriptor via `open()` which you pass to mmap. That fd came from using the filesystem driver, not directly reading bytes from the underlying device. When you write bytes via mmap, if the filesystem needs to change metadata, that is its job.

> This api is very similar to [fiemaps](https://www.kernel.org/doc/Documentation/filesystems/fiemap.txt) in linux. It gives me direct access to extents of a file, without changing the structure of the filesystem.

I am trying to understand how that works. From what I see of the linked docs, it is up to the filesystem to implement it, meaning, like mmap, it is mediated by the filesystem.

malt3:
You could rearchitect this feature to return a custom type that implements `io.ReaderAt` or `io.WriterAt` and mediate any access. I see how this doesn’t quite map to regular usage pattern for filesystems - I just find it to be very useful.
In the end, I also understand if this feature doesn’t make sense in the context of this library. I’m happy to maintain my little downstream patch if needed.

deitch:
> implement `io.ReaderAt` or `io.WriterAt`

That is quite a good idea. `os.File` supports it, no reason this should not. I would be happy to see a PR on it.

> I see how this doesn’t quite map to regular usage pattern for filesystems - I just find it to be very useful.

That explains my confusion.

I don't see exposing something that makes it too easy for people to bypass normal filesystem usage and break the filesystem. If there is a generic exposition that works, definitely.

Issue #9, opened 2019-02-18
ext4 support


the-maldridge:
This looks like a pretty amazing library and looks like an ideal package to form the basis of my installer for @void-linux.  I'd need ext4 though to do this (and someday I'd love to see support for more exotic filesystems, even it if meant shelling out to their tools.

Do you foresee ext4 landing any time soon?

deitch:
Thanks @the-maldridge . I will admit that it has been quite a labour of, well, masochism. :-)

void-linux is interesting. I was motivated to start diskfs from my work on [linuxkit](https://github.com/linuxkit/linuxkit). There are only 2 places there where we need to exec out to an external program; building filesystems is one of them. These are all bits in a file/on a disk, so I said the 5 most dangerous words any engineer can say: "how hard could it be?" 

> Do you foresee ext4 landing any time soon?

I got pretty far with ext4. If I am not mistaken, reading works fully. The difficulty is writing, and particularly some of the areas that use tree structures. Rebalancing them (or even figuring out how to build them initially) got bogged down. That led to as switch to immutable filesystems first, since they are much easier to write. Hence iso (which is complete) and squashfs (which is mostly so). I do intend to return to ext4 once squashfs is done.

CanyonCasa:
Would the [gexto](https://github.com/nerd2/gexto) package be of any use in helping you along with ext4 support?

deitch:
It really might, thank you @CanyonCasa . I can look at either integrating it, or just leveraging the few parts that got blocked. MIT and Apache licenses largely are compatible, so that is good.

deitch:
It is at a state where both squashfs and ext4 are working read but not write. Time to get it completed.

probonopd:
I'd be highly interested in squashfs read and write, and in zisofs read and write.

deitch:
I as well @probonopd :-)

Just got to find hours to work on it.

Mikkelhost:
Hey, i stumbled upon your repo!
I am looking for a way to edit a raspberry pi image using go.
So i was wondering how the ext4 support is coming along? :)

deitch:
Slowly. 😆 

ext4 reading works, you can check out the [ext4 branch](https://github.com/diskfs/go-diskfs/tree/ext4). Writing was about halfway done last I saw it. I would love to give it a month and finish it up. But real life, as they say. Every now and then, someone needs it for business purposes and invests.

Mikkelhost:
[...]

PR #264, opened 2024-11-09
Extend filesystem interface

aol-nnov:
Interface is extended as a part of ext4 improvements effort
https://github.com/diskfs/go-diskfs/issues/9#issuecomment-2422384742

Also, Remove method is introduced for all supported filesystems.

@deitch should we keep `ext4.Rm()` or we can drop it alltogether infavour of newly introduced `filesystem.Remove()`?

### DIFF
--- a/filesystem/squashfs/squashfs.go
+++ b/filesystem/squashfs/squashfs.go
@@ -49,7 +49,7 @@ func (fs *FileSystem) Label() string {
 }
 
 func (fs *FileSystem) SetLabel(string) error {
-	return fmt.Errorf("SquashFS filesystem is read-only")
+	return filesystem.ErrReadonlyFilesystem
 }
 
 // Workspace get the workspace path
@@ -229,6 +229,9 @@ func Read(file util.File, size, start, blocksize int64) (*FileSystem, error) {
 	return fs, nil
 }
 
+// interface guard
+var _ filesystem.FileSystem = (*FileSystem)(nil)
+
 // Type returns the type code for the filesystem. Always returns filesystem.TypeFat32
 func (fs *FileSystem) Type() filesystem.Type {
 	return filesystem.TypeSquashfs
@@ -266,7 +269,7 @@ func (fs *FileSystem) GetCacheSize() int {
 // if readonly and not in workspace, will return an error
 func (fs *FileSystem) Mkdir(p string) error {
 	if fs.workspace == "" {
-		return fmt.Errorf("cannot write to read-only filesystem")
+		return filesystem.ErrReadonlyFilesystem
 	}
 	err := os.MkdirAll(path.Join(fs.workspace, p), 0o755)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


aol-nnov:
Oh, just stumbled upon https://github.com/diskfs/go-diskfs/pull/233! 

What do you think about adding Rename() to the interface as well?

deitch:
I rather like this. Thank you for doing it. 

> Oh, just stumbled upon https://github.com/diskfs/go-diskfs/pull/233!

Yup, although that didn't complete the interface, and you did here. So let's complete this one, and then that can just be the specific implementation.

> What do you think about adding Rename() to the interface as well?

Sure. There is a parallel in [`os.Rename()` ](https://pkg.go.dev/os#Rename), so it works well.

> should we keep ext4.Rm() or we can drop it alltogether infavour of newly introduced filesystem.Remove()?

For now, let's deprecate it. Mark it as deprecated in the comments and refer to `Remove()`, and have it just call `ext4.Remove()`. It also parallels [os.Remove()](https://pkg.go.dev/os#Remove). In the future, we can remove it entirely.

The only other comment I had was about the returned error. I will comment inline. Once that is in, I will kick off CI and this can run and get merged in.

Thanks @aol-nnov !

filesystem/squashfs/squashfs.go
```diff
@@ -379,6 +398,13 @@ func (fs *FileSystem) OpenFile(p string, flag int) (filesystem.File, error) {
 	return f, nil
 }
 
+func (fs *FileSystem) Remove(p string) error {
+	if fs.workspace == "" {
+		return filesystem.ErrNotSupported
```
deitch, line 403: Is this the right error here? In other places, if the filesystem _supports_ the function, but not _now_ because it has become read-only, e.g. [here](https://github.com/diskfs/go-diskfs/pull/264/files#diff-c364a90e39710b015f926bf542359a918c7621bda98f597f17c53fded896918bR357-R359) we just returned:

```go
			return nil, fmt.Errorf("cannot write to read-only filesystem")
```

Granted, a specific error struct is better. 

How about create another error like you did [here](https://github.com/diskfs/go-diskfs/pull/264/files#diff-11a47696192b4602f7e0fea19870d8e2bc3be8065e259df79376c42e27a2a0bfR11-R12), something like `ErrReadOnlyFilesystem`, and then use that?

deitch (changes requested):

filesystem/squashfs/squashfs.go
aol-nnov: ErrReadOnlyFilesystem is even better! Will do it that way soon.

aol-nnov:
Also, what do you think about adding chown and chmod as well? It seems, the interface will be complete then.

deitch:
Good call.

aol-nnov:
Ok, will mock something as soon as I get to my laptop.

deitch (dismissed):

deitch:
This is great. Let's let CI go, and if green, we can merge it in.

aol-nnov:
Bummer!... Linter does not like parameters in unimplemented functions, but I'd rather keep them for documentation reasons...

Is there any annotation for such cases?

deitch:
Some of the lint errors are duplication, e.g.:

```go
func (fs *FileSystem) Link(oldpath string, newpath string) error {
```

Should be:

```go
func (fs *FileSystem) Link(oldpath, newpath string) error {
```

Some of the others are shadow imports, like:

```go
func (fs *FileSystem) Mknod(path string, mode uint32, dev int) error {
```

Which shadows the `path` package, so change the name of the parameter to something non-duplicate like `name` or `p`, etc.

For the last, unused parameter, just replace it with `_`, like:

```go
func (fs *FileSystem) Rename(_, newpath string) error {
```

Or, if you want to keep it, you can put in a nolint for it. E,g,:

```go
// Rename rename a file from oldpath to newpath
//
//nolint:unused-parameter // keep it around to make it easier in the future
func (fs *FileSystem) Rename(oldpath, newpath string) error {
```

aol-nnov:
Thank you for the hints! Will fix this up shortly!

aol-nnov:
For all the `ErrNotSupported` cases I've replaced parameter names with `_` placeholder.

As for `ErrNotImplemented` cases, unfortunately, I had no luck with `//nolint:unused-parameter`, but I've worked it around the other way:

```go
func (fs *FileSystem) Remove(_ /*pathname*/ string) error {
	return filesystem.ErrNotImplemented
}
```

The coder to implement this will then remove the placeholder and uncomment the actual parameter name.

deitch:
> I had no luck with //nolint:unused-parameter

My mistake, it should be `//nolint:revive`. See [this func](https://github.com/diskfs/go-diskfs/blob/f66ea89a8100199ca22395f32c4d7425e11040bb/filesystem/ext4/extent.go#L615-L618) for example.

CI looks clean. Let's replace that, and then we can merge it in.

aol-nnov:
Okay, done.

deitch (approved):

PR #265, opened 2024-11-09
Allows to open disk from already existing fs.File descriptor

aol-nnov:
I.e. in some `fs.FS` compliant VFS, given that `vfs.Open("some.img")` returns `fs.File`

@deitch if you have no objections, would you mind merging this improvement as well?


deitch:
It looks like this is identical to `diskfs.Open()`, except that it passes an `fs.File` rather than a `string` path to the device or file to open?

I didn't understand the need for it.

> in some fs.FS compliant VFS, given that vfs.Open("some.img") returns fs.File

If that is the case, why `OpenFile(f fs.File)` and not `OpenFile(f *os.File)`? In the end, `Open()` [calls `os.OpenFile()`](https://github.com/diskfs/go-diskfs/pull/265/files#diff-9c737fb97131b341e11a4a47cb8eaf0c6cd1fd2ecf5c33bb2ae934e346c7eaf2R328) to get an `*os.File` anyways.

aol-nnov:
> In the end, Open() calls os.OpenFile() to get an *os.File anyways.

But, actually, any implementation will fit as long as it supports fs.File + [util.File](https://github.com/diskfs/go-diskfs/blob/master/util/file.go#L9), right? To be frank, I had a thought to replace [this field](https://github.com/diskfs/go-diskfs/blob/master/disk/disk.go#L25) with interface, too, but spent too little time elaborating on it. So, to start with I decided to add just diskfs.OpenFile()

> why OpenFile(f fs.File) and not OpenFile(f *os.File)?

the only intent to use fs.File as argument is to hide type assertion inside `OpenFile()` implementation. If you insist, I can live with the assertion on my side...

deitch:
> But, actually, any implementation will fit as long as it supports fs.File + [util.File](https://github.com/diskfs/go-diskfs/blob/master/util/file.go#L9), right?

Not quite. I almost responded that anything that supports `ReaderAt`, `WriterAt` and `Closer`, but that is not quite correct.

If you look at [`initDisk`](https://github.com/diskfs/go-diskfs/blob/f66ea89a8100199ca22395f32c4d7425e11040bb/diskfs.go#L188), it actually does an `f.Stat()` in order to get the size, determine if it is a file or block device, etc.

However, `Stat()` also is a [property of `fs.File`](https://pkg.go.dev/io/fs#File), so that might work. On the other hand, `fs.File` explicitly is read-only. I once did a whole implementation of "writable FS" for an OSS project.

If you want to see if it is possible to replace `Open(f *os.File)` with `Open(f SomeFIleType)` where:

```go
type SomeFIleType interface {
    fs.File
    ReaderAt
    WriterAt
    Closer
}
```

(but much better naming), then it is worth a try. But I wouldn't break `Open()`, instead have some other function that calls it, and unify them.

aol-nnov:
> But I wouldn't break Open()

Definitely! That's why I came up with a new function for that.

>  to replace Open(f *os.File) with Open(f SomeFIleType)

I'd rather keep `SomeFIleType` internal and have public api as I proposed in the PR: `OpenFile(f fs.File, ...)` and assert type internally (and return an error if incoming type does not fit). That builds more smooth API as far as I can tell.

To cut the long story short, my typical usage is:

```go
myVfs := magicFs.New(...)
img, err:= myVfs.Open("disk.img")
disk, err:= diskfs.OpenFile(img)
```
That's why I want `diskfs.OpenFile()` accept `fs.File` as input.

I will investigate more and get back to you with my findings.

deitch:
The problem is that `fs.File` is read only, so it doesn't work. If you really need to use just fs.File, probably best off casting it outside of the Open call. 

I like the idea of an interface rather than an fs.File, but it has to have the necessary functions.

aol-nnov:
> fs.File is read only, so it doesn't work

But it works ~~on my machine~~ in my scenario )) (and, yes, to be frank, my scenario is for reading only - I didn't check any writing functionality)

Okay, I've got your points. I need some time to think out the ways...

aol-nnov:
@deitch I had some mocking attempts during the day, but it appeared a rabbit hole to replace os.File with interface, I must admit.

At some point I thought it could be worth having a read-only backing file interface for read operations and rw one for write operations.

Again, why do I stick to fs.File as the base: it is the common interface for all cases which we could type-assert the underlying storage for our needs.

So, I've created a storage class that has methods to type-assert to a required functionality and used it all over the place. Changes seem massive and not yet complete.

I just wanted to ask, if it's the right investment.. (please see my brain dump that I've offloaded today into this PR)
If not, I'd rather stick to my initial initiative with your corrections - add a single method `diskfs.OpenFile()` with the following:

```go
func OpenFile(f *os.File, opts ...OpenOpt) (*disk.Disk, error) {

	opt := &openOpts{
		mode:       ReadOnly,
		sectorSize: SectorSizeDefault,
	}

	for _, o := range opts {
		if err := o(opt); err != nil {
			return nil, err
		}
	}

	return initDisk(f, opt.mode, opt.sectorSize)
}
```

and perform all needed assertions outside. I'd be more than happy with that.

deitch:
I am trying to wrap my head around what you did. I think I see something interesting, but I am not sure. Let's see if I get it.

* `diskfs.Open()` remains unchanged
* `diskfs.OpenFile()` is new. It takes as first parameter `fs.File`, and passes it to `initDisk()`
* `initDisk()` takes an `fs.File` as first parameter, instead of `*os.File`. 
* Rather than storing `Disk.Writable` as a separate property, it has a new structure called `storage.Disk` that wraps the underlying disk device, as well as storing the writable state
* Whenever we otherwise would have tried to write to the device, rather than checking `d.Writeable`, we call `d.File.Rw()`. If it returns `RwFile`, we know it is read-write, and then we can use it, otherwise return an error.  For things where we do not need to read, we can call `d.File.Ro()` and get `RoFile`.
* Lots of underlying functions get passed this `RoFile` or `RwFile` rather than whatever they had before, I guess for stricter type-checking.

The goals appear to be:

* Have the ability to pass an `fs.File`, which is more universal than `os.File`
* Have the ability to pass an `fs.File`, a read-only file structure, when we only will do read operations

Is that about it? If so, I think I like the goals, and most of the direction. I would suggest some simplifications, but let's wait until I see if I understood you correctly.

deitch:
OK, got it. I have some thoughts on simplifications. They might or might not be worth it, once I type them out. I will get to them a bit later.

aol-nnov:
Whenever you're comfortable to! No rush. I'd be glad to hear your suggestions! To be frank, that massive plot twist during my coding session was unexpected, but yet, I decided to share my thoughts ))

 I'm leaving for today (I'm GMT+3) and will be back online in next 12 hours or so. I'll have occasional access from smartphone for next 4 hours, though.

deitch:
I will comment here, although some of it may be inline as well.

### devcontainer

I don't mind have `devcontainer.json` for it, but it should be in a different PR. I also would likely use the standard `library/golang`.

### storage.File

Originally, I didn't like the idea of it being in a new package, but I have come around to it.

The language of `Ro` vs `Rw` is somewhat misleading. All underlying storage at least must support read; some functions require it to be write as well. So maybe `storage.File` should be readable as is, while you can get a "writable variant" if you ask for one. Something like (I know this isn't 100% correct, just the idea):

```go
type File struct {
	fs.File
	io.ReaderAt
	io.Seeker
	io.Closer
}


type Writable interface {
	File
	io.WriterAt
}

// OS-stecific file for ioctl calls via fd
func (f File) Os() (*os.File, error) {
	if osFile, ok := f.storage.(*os.File); ok {
		return osFile, nil
	}
	return nil, errNotSuitable
}

// file for read-only operations is not needed, by definition I can use `storage.File` for anything readable

// file for writable operations
func (f File) Writable() (Writable, error) {
	if rwFile, ok := f.storage.(Writable); ok {
		if !f.readonly {
			return rwFile, nil
		} else {
			return nil, errIncorrectOpenMode
		}
	}
	return nil, errNotSuitable
}
```

So my usage becomes:

```go
func (d *Disk) ReadPartitionContents(part int, writer io.Writer) (int64, error) {
        /* this all goes away, since d.File _is_ readable
	backingRoFile, err := d.File.Ro()
	if err != nil {
		return -1, err
	}
        */
...
	return partitions[part-1].ReadContents(d.File, writer)
```

### usage of struct vs interface

I think this only appears in one place, but changing this:

```go
func (fs *FileSystem) readMetadata(r io.ReaderAt, c Compressor, firstBlock int64, initialBlockOffset uint32, byteOffset uint16, size int) ([]byte, error) {
```

to that:

```go
func (fs *FileSystem) readMetadata(r storage.RoFile, c Compressor, firstBlock int64, initialBlockOffset uint32, byteOffset uint16, size int) ([]byte, error) {
```

doesn't really buy anything. The only part of `storage.RoFile` (or just `storage.File`) that `readMetadata()` needs is its `io.ReaderAt`, so might as well leave it as the most basic interface needed.

### initDisk

I like what you did there. Why pass it an `*os.File` when a more abstract `fs.File` will do. Since `os.File` implements `fs.File`, it can be called in the same way without changing the calling function. Much cleaner. Use the most basic interface needed (similar to the point in the previous section).

Thoughts?

aol-nnov:
Thank you for the detailed comments, Avi!

Devcontainer configs (and I've added some vscode configs later, too) definitely deserve a separate PR. Will split that into separate one in a moment. As for container image, it was offered by vscode wizard, so I suppose it contains some vscode-related tweaks for more comfortable development. I didn't dig any further in that direction yet.

As for newly introduced `storage` package, it was needed to avoid import cycles. At start I was trying to avoid it as much as I could, but failed.

I've got your idea about `storage.File` improvements and will proceed in that direction, thank you.

Regarding usage of struct vs interface in `readMetadata` and such, totally agreed. I will revert that part. It was a collateral damage of my fight with red underlines (type mismatches) all over the place during my intial attempt. Did not think much during that fighting, ha )

aol-nnov:
Hey, @deitch ! Had some time to implement your suggestions today. Please, check if I'm moving in the right direction and got everything right.

Tests are still failing, but I'm on it. Just syncing and showing you my progress.

deitch:
Yeah, that is pretty nice. Cleaner and easier to understand. It took more work, but I like it.

aol-nnov:
Glad you like it!

I'll fix the tests then and we'll see it in action.

Btw, while separating ro and rw operations I've stumbled upon a fat32 test, where traversing directories actually was creating them. Now it's failing and I temporary commented it out. Needs more thinking! ))

deitch:
> Btw, while separating ro and rw operations I've stumbled upon a fat32 test, where traversing directories actually was creating them. Now it's failing and I temporary commented it out. Needs more thinking! 

We have had cases like that before. Do you mind opening a separate issue just around the fat32 test and how to recreate it?

aol-nnov:
> Do you mind opening a separate issue just around the fat32 test

(putting in my to-do list) Will do, eventually! ))) I haven't yet dove into concrete implementations yet.

aol-nnov:
Oh, wow...

To my surprise, the similar plot is rolling out on the qcow2 branch! Just noticed it and now the puzzle fits quite well, yay!

May be we should tweak the naming here, too? It's no longer a `d.File`, it's something more generic!

It's now something like a "backend storage". `Backend`? Or, may be a `Driver` as you named it those days?

deitch:
> the similar plot is rolling out on the qcow2 branch! Just noticed it and now the puzzle fits quite well, yay

Do you mean the test directory duplication issue?

> It's now something like a "backend storage". Backend? Or, may be a Driver as you named it those days?

No objection. I don't really think it is a driver, as it doesn't include lots of logic. It is a storage or backing store or backend.

aol-nnov:
> Do you mean the test directory duplication issue?

No, I mean the whole interface-instead-of-`os.File` thing! ))

aol-nnov:
Once in a while I'm thinking to myself, isn't it a time for v2, ha! ))

deitch:
> P.S.: I don't get it.. Why does it fail only on ubuntu?! Darwin has the same tests, no?!

Because the test step [only runs on ubuntu](https://github.com/diskfs/go-diskfs/actions/runs/11871313640/workflow?pr=265#L70-L72):

```yaml
      - name: test
        if: matrix.os == 'ubuntu-latest'
        run: make test
```

deitch:
Actually, this is looking pretty good. I do not know the issue with fat32 mkdir, but it looks like it has to do with the FSIS:

```
failed to write the file system information sector: disk file or device not open for write
```

Is it possible that the writablefile is not being passed through somewhere, or that writing FSIS is using just File?

aol-nnov:
> Is it possible that the writablefile is not being passed through

Yeah, that was the case!

Phew, seems, we're on the homestretch with this change!

Should we add some sugar like

```go
disk, err := diskfs.WithBackend(raw.New(...))
```

Also, should we add any `Resize()` or `Create()` to `backend.Storage` interface?

Btw, I do not mind if you merge fat32 improvements first. I'll rebase on top of that.

deitch:
Heh, already did merge those fat32 improvements. 😁

aol-nnov:
Rebase was super smooth - no conflicts!

deitch:
How is this looking from your perspective? I see it still is draft, but all checks pass.

> Also, should we add any Resize() or Create() to backend.Storage interface?

I don't think so. Resize can only happen in certain circumstances, and it is messy, I don't want us to complicate things.

> Should we add some sugar like `disk, err := diskfs.WithBackend(raw.New(...))
> instead of newly introduced diskfs.OpenFile()?

Usually `With*()` is used for variadic functional arguments, like:

```go
foo.Create(foo.WithOpt1(), foo.WithSomeOtherOpt(), etc.)
```

We actually do that for [OpenOpt](https://pkg.go.dev/github.com/diskfs/go-diskfs#OpenOpt), so not `diskfs.WithBackend()`, but if you are thinking the same idea with different naming, maybe?

Right now, you get a disk via one of:

* `diskfs.Create(pathToFileOrDevice)`
* `diskfs.Open(pathToFileOrDevice)`

It intentionally mirrors [os.Create()](https://pkg.go.dev/os#Create) and [os.Open()](https://pkg.go.dev/os#Open), although I think we have shown the limits of that. There is nothing parallel to `Storage` in `os`, which isn't surprising.

Maybe `diskfs.OpenStorage()`?

aol-nnov:
> How is this looking from your perspective?

I think, it's ready overall, just final touches left that I've pointed in my previous comment.

> Resize can only happen in certain circumstances, and it is messy

Actually, by introducing `Resize()` I was trying to address [this](https://github.com/diskfs/go-diskfs/blob/master/examples/create-iso-from-folder/main.go#L40) and similar situations.

Ideally, I was planning to move away from `diskfs.Open()`/`diskfs.Create()` and offload this operations to the `backend` implementation completely. (That's why I was talking about my v2 thoughts recently)

> Maybe `diskfs.OpenStorage()`?

Well, seems legit if you're not ready to drop current `diskfs.Open()`/`diskfs.Create()` functions.

Also, I've made attempt to revive qcow2 branch (although, I do not know much about that format) by snitching some code from that branch and making it fit newly introduced interface. Please [take a look](https://github.com/aol-nnov/go-diskfs/tree/revive-qcow2). I'd really like to keep the original authorship of that code, so, if you like it, could you refresh that branch yourself?

deitch:
> Actually, by introducing Resize() I was trying to address [this](https://github.com/diskfs/go-diskfs/blob/master/examples/create-iso-from-folder/main.go#L40) and similar situations

I get it. In that case, let's get this one in, as it is a big change. Whether or not we want to support resize can be its own conversation in a separate issue or PR. Make sense? 

> Ideally, I was planning to move away from diskfs.Open()/diskfs.Create() and offload this operations to the backend implementation completely. (That's why I was talking about my v2 thoughts recently)
> Well, seems legit if you're not ready to drop current diskfs.Open()/diskfs.Create() functions

So what would it look like then? diskfs always will have `Open` and `Create`, to distinguish between "open an existing filesystem" on this storage, and "create a new filesystem" on this storage. Difference between `mount` and `mkfs`, essentially.

> Also, I've made attempt to revive qcow2 branch (although, I do not know much about that format) by snitching some code from there and making it fit newly introduced interface. Please [take a look](https://github.com/aol-nnov/go-diskfs/tree/revive-qcow2). I'd really like to keep the original authorship of that code, so, if you like it, could you refresh that branch yourself

Yeah, that one slipped. ext4 became more important, and got the final push through. It still is missing some functionality, but moved ahead a lot, enough to be functional.

Let's look at qcow2 as a separate question. I won't 100% have time for that for a bit.

deitch:
The nice thing about this repo is that the tests run fairly quickly (unlike some other projects with which I am involved). So separating work into multiple PRs is low overhead.

aol-nnov:
> Whether or not we want to support resize can be its own conversation in a separate issue or PR

Totally agreed. 

>  diskfs always will have Open and Create, to distinguish between "open an existing filesystem" on this storage, and "create a new filesystem" on this storage.

Controversial situation, indeed. I was [trying to introduce](https://github.com/aol-nnov/go-diskfs/blob/open-file/backend/raw/raw.go#L28) some sort of "constructor" for "creating a new image" as opposed to "constructor" for [existing device/image](https://github.com/aol-nnov/go-diskfs/blob/open-file/backend/raw/raw.go#L18) but I must admit, it's still not that expressive as it should be...

As for qcow2, I'm not pushing you in that direction by any means - just noticed the stalled branch and decided to check if it actually fits new reality, ha!

deitch:
> Yeah! I'm really enjoying the speed!
> In some projects it takes from 30 minutes to hours to complete the PR pipeline - so frustrating!...

If you _really_ want and think it is important, I am sure I could come up with some way to make the CI pipeline take hours.

aol-nnov:
>  I could come up with some way to make the CI pipeline take hours.

Hope, you're kidding me now, Avi!.. ))

aol-nnov:
So, `OpenStorage` it is or are you planning to give more comments on [that](https://github.com/diskfs/go-diskfs/pull/265#issuecomment-2482338438)?

aol-nnov:
What a harsh linter! Does not allow to use deprecated methods, ha!.. Well, okay, I'll soften the wordings....

aol-nnov:
>  Is there a parallel for Create?

Not more than `disk.OpenStorage(raw.CreateFromPath("..."))`, as I stated [earlier](https://github.com/diskfs/go-diskfs/pull/265#issuecomment-2482338438).

If you have better ideas, please let me know!

deitch:
> Not more than disk.OpenStorage(raw.CreateFromPath("...")), as I stated https://github.com/diskfs/go-diskfs/pull/265#issuecomment-2482338438.

Are you saying that `diskfs.Create()` is not really needed? [Here is the source](https://github.com/diskfs/go-diskfs/blob/v1.4.2/diskfs.go#L305-L355) for `Open()` and `Create()`:

```go
func Open(device string, opts ...OpenOpt) (*disk.Disk, error) {
	err := checkDevice(device)
	if err != nil {
		return nil, err
	}

	opt := openOptsDefaults()
	for _, o := range opts {
		if err := o(opt); err != nil {
			return nil, err
		}
	}

	m, ok := openModeOptions[opt.mode]
	if !ok {
		return nil, errors.New("unsupported file open mode")
	}

	f, err := os.OpenFile(device, m, 0o600)
	if err != nil {
		return nil, fmt.Errorf("could not open device %s with mode %v: %w", device, m, err)
	}
	// return our disk
	return initDisk(f, ReadWriteExclusive, opt.sectorSize)
}

// Create a Disk from a path to a device
// Should pass a path to a block device e.g. /dev/sda or a path to a file /tmp/foo.img
// The provided device must not exist at the time you call Create()
func Create(device string, size int64, _ Format, sectorSize SectorSize) (*disk.Disk, error) {
	if device == "" {
		return nil, errors.New("must pass device name")
	}
	if size <= 0 {
		return nil, errors.New("must pass valid device size to create")
	}
	f, err := os.OpenFile(device, os.O_RDWR|os.O_EXCL|os.O_CREATE, 0o666)
	if err != nil {
		return nil, fmt.Errorf("could not create device %s: %w", device, err)
	}
	err = os.Truncate(device, size)
	if err != nil {
		return nil, fmt.Errorf("could not expand device %s to size %d: %w", device, size, err)
	}
	// return our disk
	return initDisk(f, ReadWriteExclusive, sectorSize)
}
```

Are you saying that we don't really need `Create()`, it could be moved to `raw.CreateFromPath()`, because all it does is:

1. ensure that a path was provided - storage takes care of that
2. Open the file with `RDWR|EXCL|CREATE`, so it creates the file if it does not exist
3. Ensures the size is provided and then grow/shrink the file to that size
4. call `initDisk`

So you are saying, that should be outside of `diskfs` basic functions, part of storage - create a file, set its size - and then you just would call `Open()` like any other that exists?

aol-nnov:
You caught my thought perfectly correct!

That's what I was trying to say while introducing `Resize()` and `Create()` [here](https://github.com/diskfs/go-diskfs/pull/265#issuecomment-2481322751).

But definitely, it all needs more thinking... If you're fine with merging this PR as it is now and proceed in the next PRs, I'm totally fine, too!

deitch:
> Also, this change will introduce the API breakage, so we need to decide how to handle such a big change

This one? This doesn't break anything. 😕 
Oh, maybe `Create()`? Why break that?

> You caught my thought perfectly correct!

Then we should document it. Maybe add an example to show how to use the storage to Create or Open, and the main doc a bit?

aol-nnov:
> This one? 

No, the one you proposed!

> and then you just would call Open() like any other that exists?

In such case we could introduce a single `diskfs.Open(b storage.Backend)` but golang does not support method overloading as C does. Thus the API breakage - `diskfs.Open` would have a totally different set of parameters..

aol-nnov:
> Maybe add an example to show how to use the storage to Create or Open, and the main doc a bit?

Yeah, will do later today or tomorrow morning! Also, I've got an idea to add `raw.OpenFromPath("..")` to parallel `diskfs.Open`.

So, making it as draft again! ))

aol-nnov:
Added some example usage and `raw.CreateFromPath()`/`raw.OpenFromPath()` utility methods.

aol-nnov:
@deitch probably, I had to make it more clear, that I was done with it - too much 'draft'-toggling back and forth... So, this is it, finally, ha!

Please review when you have time and let me know if there are any more improvements needed.

deitch:
Sorry, took me some time to get to it. Trying now.

README.md
```diff
@@ -16,10 +16,18 @@ Note: detailed go documentation is available at [godoc.org](https://godoc.org/gi
 ### Concepts
 `go-diskfs` has a few basic concepts:
 
+* Backend
 * Disk
 * Partition
 * Filesystem
 
+#### Backend
+Backend is a (relatively) thin layer which abstracts low-level read/write operations. Through a backend you can seamlessly operate different disk formats (see a bit stale qcow2 branch).
```
deitch, line 25: Please get rid of the "bit stale qcow2 branch" comment. It is fine in a comment on a PR or issue, but not in long-lived docs.

deitch: Also, we might want to make it clear that you can use a Backend or just use a disk directly?

backend/raw/raw.go
```diff
@@ -0,0 +1,127 @@
...
+	return rawBackend{
+		storage:  f,
+		readonly: isReadonly,
+	}
+}
+
+// Create a backend.Storage from a path to a device
+// Should pass a path to a block device e.g. /dev/sda or a path to a file /tmp/foo.img
+// The provided device/file must exist at the time you call OpenFromPath()
+func OpenFromPath(pathName string, isReadonly bool) (backend.Storage, error) {
```
deitch, line 29: I think you can just name this `readOnly`, which is fairly common. 

Actually, now that I think about it, do we want to just do `Open()` and have mode, paralleling `os.Open()`? That would give us the ability to have `Open()` even create the file. Or do we think this is simpler?

backend/raw/raw.go
```diff
@@ -0,0 +1,127 @@
...
+	return rawBackend{
+		storage:  f,
+		readonly: isReadonly,
+	}, nil
+}
+
+// Create a backend.Storage from a path to an image file.
+// Should pass a path to a file /tmp/foo.img
+// The provided file must not exist at the time you call CreateFromPath()
+func CreateFromPath(pathName string, size int64) (backend.Storage, error) {
```
deitch, line 58: See the comment above. If we have `Open()` with mode, then `Create()` does not need to exist, or at most is a wrapper around `Open(os.O_CREATE)`. Or do we think it is simpler this way?

README.md
```diff
@@ -16,10 +16,18 @@ Note: detailed go documentation is available at [godoc.org](https://godoc.org/gi
...
+
+Currently there is only one implementation - raw.
+
+Use `raw` backend to access block devices and raw image files.
```
deitch, line 29: Reading this twice made me think, is "raw" the right name? Or should the backend just be `file` or `path`? That would simplify the interface to `file.Open()` and `file.Create()` (or `path.Open()` and `path.Create()`).

backend/interface.go
```diff
@@ -0,0 +1,33 @@
...
+	fs.File
+	io.ReaderAt
+	io.Seeker
+	io.Closer
+}
+
+type WritableFile interface {
+	File
+	io.WriterAt
+}
```
deitch, line 25: I like how you defined the interfaces clearly here. Thank you.

disk/disk.go
```diff
@@ -68,24 +60,19 @@ func (d *Disk) GetPartitionTable() (partition.Table, error) {
...
 	// fill in the uuid
-	err := table.Write(d.File, d.Size)
+	err = table.Write(rwBackingFile, d.Size)
 	if err != nil {
 		return fmt.Errorf("failed to write partition table: %v", err)
 	}
 	d.Table = table
-	// the partition table needs to be re-read only if
-	// the disk file is an actual block device
-	if d.Type == Device {
```
deitch, line 82: Why do we always reread the partition table here? I thought we only have to if it is a device. Or did you move that logic into the backend somewhere?

deitch (commented):
Mostly some questions, a few recommended naming changes. Overall, this looks solid. Thank you.

aol-nnov:
> Sorry, took me some time to get to it. Trying now.

No worries and, definitely, no rush!

I'll address your comments tomorrow. I'm a bit overwhelmed with U-Boot back-porting right now, ha ))

backend/raw/raw.go
aol-nnov: > and have mode, paralleling os.Open()

no other mode bits except os.O_RDONLY are configurable by user here, so I opted to a boolean flag. How could we need all mode bits (as seen in os.Open()) here?

backend/raw/raw.go
aol-nnov: I've separated them intentionally to not overwrite existing image. Open just opens and Create, ahem, creates! )) In no chance existing file will be truncated. But I agree, it looks like some sugar around lower level functions.

What would be your final thoughts on this?

README.md
aol-nnov: > Reading this twice made me think, is "raw" the right name?

I was thinking of naming backends according to supported image formats. So, raw backend supports raw images. The next one will be, probably, qcow2.

Do you think, it is misleading?

backend/interface.go
aol-nnov: Glad you liked it!

disk/disk.go
aol-nnov: Correct! The logic was moved inside the function itself, as it works only with device files. https://github.com/aol-nnov/go-diskfs/blob/open-file/disk/disk_unix.go#L29

disk/disk_unix.go
```diff
@@ -18,10 +19,24 @@ const (
 //
 // It is done via an ioctl call with request as BLKRRPART.
 func (d *Disk) ReReadPartitionTable() error {
-	fd := d.File.Fd()
-	_, err := unix.IoctlGetInt(int(fd), blkrrpart)
+	// the partition table needs to be re-read only if
+	// the disk file is an actual block device
+	devInfo, err := d.Backend.Stat()
```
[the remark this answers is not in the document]
aol-nnov: No change at all, actually. Just moved the wrapping `if-s inside the function (see [previous comment](https://github.com/diskfs/go-diskfs/pull/265#discussion_r1864210298))

util/backend.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"io/fs"
+
+	"github.com/diskfs/go-diskfs/backend"
+	"github.com/diskfs/go-diskfs/backend/raw"
+)
+
+// Returns backend.Storage or creates Raw backend around provided fs.File
+func BackendFromFile(f fs.File, readOnly bool) backend.Storage {
```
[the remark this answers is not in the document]
aol-nnov: Well, another controversial part of this changeset, ha!

Actually, this method checks if provided fs.File looks like a backend (any suitable implementation) and if it is, then passes on it as is. So based on that thought I did not include it into raw backend implementation. It is a utility method for all backend.

Initially I'd like to put it into `backend` package along with interface definitions, but it appeared to be impossible due to circular imports...

README.md
aol-nnov: Well, we could add some notice like this in the backend section of readme:

> For convenience, if provided fs.File is not `backend`-like, it is internally wrapped into `raw` backend. But keep in mind, this will work with device files and raw images only! Opt to creating backend instance explicitely for the clarity sake.

deitch: > Not exactly - it stands true only in the case of using raw images or device files...
> If provided fs.File is not a backend-like, it is internally wrapped into raw backend, but that provided fs.File might be not a raw image or device and in that case the user will eventually end up with an error.
> So, I'd propose using backends explicitly for clarity.

I don't understand. There still are the `Open(path string)` and `Create(path string)` functions, so you still can use a file directly? It is a convenience function equivalent to (pseudo-code) `Open(backed.FromFile(path)`

deitch: > Well, we could add some notice like this in the backend section of readme

I didn't quite understand this either. What does "backend-like" mean, or internally wrapped?

backend/raw/raw.go
deitch: > no other mode bits except os.O_RDONLY are configurable by user here

Don't you have the equivalent of `Create()`, which would be `Open(path, os.O_CREATE|os.O_RDWR)`?

Again, I myself am kind of torn on this. I know that logical OR of flags is the way Unix-like has done it for over 50 years, but I never really liked it either.

backend/raw/raw.go
deitch: As you probably can tell by now, my lower-level thoughts are mutually contradictory. 😆

I think for now you are right, we should leave it with two distinct functions: `Open()` and `Create()`.

README.md
deitch: But qcow2 is a filesystem, which is lower-level than a disk. You get to qcow2 by doing `diskfs.Open()` (or `Create()`), which returns a handle to a `Disk`. What you do with that - reading a partition table, or a filesystem, or creating one, or partition with a filesystem on one of the partitions, is all at a next level down. `Disk` doesn't care or know about what the internal structure is.

diskfs.go
```diff
@@ -240,24 +238,32 @@ func Open(device string, opts ...OpenOpt) (*disk.Disk, error) {
 	return initDisk(f, ReadWriteExclusive, opt.sectorSize)
 }
 
+// Open a Disk using provided fs.File to a device in read-only mode
+// Use OpenOpt to control options, such as sector size or open mode.
+func OpenStorage(f fs.File, opts ...OpenOpt) (*disk.Disk, error) {
```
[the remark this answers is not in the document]
deitch: > OpenStorage is what you advised me during some chat

Yeah, well, I never said I had all the right answers! 😁 

Working it through with you has made some of this clearer.

You are right, ideally it would be `Open(f fs.File)`, and then you would have a convenience function like `OpenFile(device string)` (or `OpenPath()`, etc.). But I am hesitant to break backwards-compatibility unless it brings a _lot_ of benefit.

Well, everything else became `Backend`, so maybe `OpenBackend()`?

util/backend.go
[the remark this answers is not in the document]
deitch: If you are making a backend from a file, and that backend is identical to what you would get from the package currently named `raw`, then it goes there. It fits better if this is called `file` or `path` anyways.

README.md
aol-nnov: > What does "backend-like" mean

A type, that fully implement [backend.Storage](https://github.com/aol-nnov/go-diskfs/blob/open-file/backend/interface.go#L27) interface

>  or internally wrapped?

Here it is: https://github.com/aol-nnov/go-diskfs/blob/open-file/diskfs.go#L117-L121

Seems, I've chosen not the best terms for explanation, sorry..

backend/raw/raw.go
aol-nnov: So, let's keep the boolean flag here?

diskfs.go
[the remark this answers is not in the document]
aol-nnov: Let it be `OpenBackend` for now, okay!

util/backend.go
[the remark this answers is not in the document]
aol-nnov: Well, I've made another round on that function and it seemed to me, it is unnecessary now. So, I've removed it completely now.

diskfs.go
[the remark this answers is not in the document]
aol-nnov: > Yeah, well, I never said I had all the right answers! 😁

Absolutely! ))

I trust the process and believe, we'll work out the best solution together, iteratively!

README.md
aol-nnov: > But qcow2 is a filesystem, which is lower-level than a disk

Correct, it is on the same level as raw backend is!

> You get to qcow2 by doing diskfs.Open()

Something like `diskfs.OpenBackend(qcow2.New("./the_image"))`, right? Disk does not care or know about what the internal structure is, indeed!

backend/interface.go
```diff
@@ -0,0 +1,33 @@
...
+
+type Storage interface {
+	File
+	// OS-stecific file for ioctl calls via fd
```
deitch, line 29: Typo: "OS-specific"

README.md
deitch: >> What does "backend-like" mean
> A type, that fully implements [backend.Storage](https://github.com/aol-nnov/go-diskfs/blob/open-file/backend/interface.go#L27) interface

Haha. OK, now I get it. In that case, you mean, "if the argument implements the `backend.Storage` interface, then..."

There are only several possibilities:

* call `OpenBackend(f fs.File)`
* call `Open(p string)` - this only accepts a path to a device or disk image
* call `Create(p string)` - this only accepts a path to a device or disk image

The latter two cases are just convenience wrappers around `OpenBackend(backend.FromFile(p))` (pseudo-code), so you are not referring to that, you must be referring to `OpenBackend(f fs.File)`. Why would we pass something that needs to be "wrapped"? I guess maybe if you have a read-only backend, e.g. I want to use an `fs.File` handle that I have that refers to some storage that has on it a disk image? I guess you are saying that `Disk` must have a `Backend` property, so you need to translate the passed `fs.File` into `Backend`?

Would that even work? Don't I need an actual `backend.File` at minimum?

README.md
deitch: Yeah, but I don't think you need a separate backend for qcow2. Backend does not know or think about the contents. 

```
---> Disk (wraps) ---> Backend (handles) ---> low-level i/o
```

Dealing with qcow2 is handled at a higher level, not at a lower level. If you look at the qcow2 branch, it just changes how `Disk` deals with its writes. Adding your (good) `Backend` abstraction shouldn't change that. It just abstracts out i/o; what gets written to it is handled in `Disk` (or a level above).

To use your example, I think it would be more `d := diskfs.OpenBackend(file.New("./the_image"))`. Inside of `d` we would have ways of handling the raw vs qcow2. I don't think it goes in `Backend`. I may discover otherwise as I pick up qcow2 again, but I don't think so.

backend/interface.go
aol-nnov: Whoopsie.. and in two different places! Copy-paste wins..

Fixed that, thank you!

README.md
aol-nnov: Well, to be frank, the more I read this, the more I dislike it... Moreover, func `OpenBackend(f fs.File..` looks somewhat contradictory to me - function name says "backend" and argument says "File", what?!

I think, it'll be more clear if we pass a backend as argument like `func OpenBackend(b backend.Storage, opts ...OpenOpt)`.

README.md
aol-nnov: Well, to be frank, I did not dug deep into qcow2 branch, but I saw that it has its own implementation of [ReadAt](https://github.com/aol-nnov/go-diskfs/blob/revive-qcow2/backend/qcow2/qcow2.go#L136) and other methods from backend.WritableFile, so I thought to myself, that it'd be a perfect match!

Nice separation of concerns - `backend` abstracts i/o and `Disk` (on the next level) handles partitions, filesystems, etc. without knowing anything about underlying storage. This way we do not introduce any backend implementation details to `Disk` at all.

I might be wrong, though..

README.md
aol-nnov: Changed it that way https://github.com/aol-nnov/go-diskfs/blob/open-file/diskfs.go#L241

What do you think, looks better, eh? )

backend/raw/raw.go
aol-nnov: So, we're leaving it as it is now?

Resolving this to reduce noise in discussion threads.

README.md
deitch: It does.

README.md
deitch: > Nice separation of concerns

Exactly. Which is why I think qcow2 is at a higher level. `Backend` is just i/o, it doesn't know if it is raw or qcow2 or something else entirely.

aol-nnov: I won't argue for long )) , but my final thoughs are this: look at how it is organised in linux - you can mount raw images with loop mount and qcow2 images, say, via fuse. No portions of kernel dedicated to partitions and filesystems are aware of qcow2 format. This knowledge is encapsulated in the lower level.

If you're still not convinced, I'll remove lines in question tomorrow, as I'm leaving for today.

I'm really not much aware of your plans regarding the future, so no pushing from my side - just thinking aloud!

deitch: > won't argue for long

Why not? It keeps life interesting. Arguing with smart people is one of the great intellectual joys of life.

I do understand your point. Mostly, in my head, I am mapping it to how we work now. That is, "here is a file path (disk image or block device), get a reference `Disk` to it, and then we can use that to get the internal structure of it."

How about we rename `raw` to `file` for now, assume it is the only way. If we have to redo it to support qcow2 in the future, as we go back to it, we can. I think it will be the same backend, maybe with different properties.

aol-nnov: > Arguing with smart people is one of the great intellectual joys of life.

This FLUTTER-s me, so I'm GO-ing to add some more ))) (all pun intended, ha!)

>  I think it will be the same backend, maybe with different properties.

This bothers me a bit, to be frank... It's a bit against the beloved "unix way", when each part performs one simple task.

What if, by the time passes, you decide to add more formats? Say, vmdk or anything else? (Disclamer: I'm not well aware of those formats internals) Everything will be placed inside that `file` backend with different properties?

To sum up, I tend to keep things separate: raw bachend as a reference implementation and more complex implementations alongside of it.

Moreover, I had some thoughts about factoring ext4, fat and iso out of this repo and to establish an API for registering external implementations to allow users to use their own.

I understand, that implementing such things in users-space is quite a niche task and only a few users would benefit from such a change, but it would lighten the core of this project, too.

Still not convinced? Then, I'll obey you as BDFL of this project ))

aol-nnov: Also, if you'd like to bring some magic (as in `man 5 magic`) to users by automatically selecting the backend based on provided file image format, we could implement such a selector, too.

deitch: > bring some magic

Do you mean something like [this magic](https://github.com/deitch/magic)? Haven't touched it in some time, but still use it in a few places. 😁 

> bit against the beloved "unix way"

I disagree, I think it is precisely the Unix way. Backend handles one thing, and one thing only: interfacing with a storage device. It is a "storage driver", not a "format driver". Following the unix way, as you suggested, I could see us splitting a "file" backend from a "block device" backend (or device driver, etc.), i.e. difference between `/home/aol-nnv/disk.img` and `/dev/sd0`. Right now, they are really close, we just have a bit of logic to distinguish between them; that would be easier if we had different backends. Another might be `memory` (for really small ones).

If we did vmdk (I looked at it at some point, really remember nothing about it), like qcow2, these would be a level above, right now in `Disk`. Also, I like that idea, although vmdk is becoming less and less interesting, as VMWare itself appears to be going through some challenges post-acquisition. On the other hand, demand for utilities to handle those may grow. Who knows?

> had some thoughts about factoring ext4, fat and iso out of this repo and to establish an API for registering external implementations to allow users to use their own

Feel free to open an iissue to discuss it; I want to see where your thoughts are going.

> Still not convinced? 

Nope, not really. I can see how in the future we might want some layer to distinguish between disk types, beyond having them part of `Disk` itself. I don't think those will be `Backend` - which will be purely i/o drivers, sort of "layer 2" (OSI model) - but either part of `Disk` or something else yet to be discovered.

Of course, we only will know for sure once we get there.

aol-nnov: > Do you mean something like this magic? 

Yep, exactly)

> What is "BDFL"?

https://en.m.wikipedia.org/wiki/Benevolent_dictator_for_life

I thought, it's an established term among programmers 😁

Okay, I'll try to find a few minutes tomorrow to finalize this. Turned into a bit of a journey already)) (and I'm not saying I didn't like it!)

aol-nnov:
@deitch  I've renamed `raw` backend to `file` as per you request. We're close to merge complete than ever before ))

deitch (approved):

deitch:
That was some job, wasn't it? Thank you for all the effort!

aol-nnov:
It's a pleasure to deal with you, Avi! Looking forward to contribute more! :)

PR #266, opened 2024-11-12
Add VS Code tooling

aol-nnov:
@deitch tooling is factored out into this PR.

Comments are welcome!

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,28 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/go
+{
+	"name": "Go",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	"image": "mcr.microsoft.com/devcontainers/go:1-1.22-bookworm",
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	// "features": {},
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [],
+	// Use 'postCreateCommand' to run commands after the container is created.
+	"postCreateCommand": "sudo apt-get update && sudo apt-get -y install dosfstools mtools gdisk fdisk p7zip genisoimage squashfs-tools && echo 'mtools_skip_check=1' | sudo tee -a /etc/mtools.conf",
+	"customizations": {
+		"vscode": {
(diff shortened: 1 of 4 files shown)


aol-nnov:
> where did you get the devcontainer configs from

VS Code did that for me. Just pressed cmd+shift+p and typed few words in! )

image

> should the .vscode/ be committed

Few files from that directory indeed should be ignored, but according to canonical gitignore, settings and extensions are allowed. https://www.toptal.com/developers/gitignore/api/visualstudiocode

Also, I'd like to share the corresponding plugin: https://marketplace.visualstudio.com/items?itemName=piotrpalarz.vscode-gitignore-generator which is extremely handy!

aol-nnov:
By the way, should we update the root `.gitignore` with the help of plugin I'm referring to?

aol-nnov:
Look, I've added the generated file. What do you think?

deitch:
> Few files from that directory indeed should be ignored, but according to canonical gitignore, settings and extensions are allowed. https://www.toptal.com/developers/gitignore/api/visualstudiocode

I do not know what toptal is, looks like a recruiting site? Although gitignore.io (which is not necessarily authoritative either) recommends the same pattern. 🤷‍♂️ 

It feels strange to me, trying to tell people what extensions they should install and what settings to use. The recommendations do make sense, though, as it is just recommendations. But should we be including vscode specific stuff in the repo?

In any case, I am not sure about all of those recommendations. For example, I only use the 3rd and 4th. If someone wants the extra, that is fine. Not sure we should be pushing them.

Some of the settings as well, although those, I think, we can live with.

aol-nnov:
> I do not know what toptal is, looks like a recruiting site?

Toptal acquired gitignore.io few years ago. All ignore patterns are [available on github](https://github.com/toptal/gitignore.io)
 and are managed by programming community. I've used gitirnore.io for more than a decade now, different languages, different IDEs. It served me well through years.

>  what extensions they should install and what settings to use

As for plugins, this is in no way an enforcement, it is a recommendation. When you open the workspace, a popup is showed informing the user about plugins recommended for comfortable development of the current project.

As for settings.json, it is also for consistency. For example, format on save, newline at the end of file - all that essentials.

Anyway, I'm open to any suggestions regarding any of the configs and, on the other side, invite you to try new vscode plugins if you haven't heard of them before. Chances are high, you'll like them!

deitch:
Nah, I think it is ok. We always can change it if it gets too rough.

deitch (approved):

aol-nnov:
Yay, cool!

PR #268, opened 2024-11-18
Create godoc-friendly examples

aol-nnov:
Move tests from the diskfs.go global comment section and update them as they are slightly out of date.

@deitch while trying to put my examples together I've found out that code snippets from the [header comment block](https://github.com/diskfs/go-diskfs/blob/master/diskfs.go#L12-L102) are slightly out of date.

I've converted them to the godoc-friendly versions and also left one `FIXME` item, as I'm not sure how it was intended to work [here](https://github.com/diskfs/go-diskfs/blob/master/diskfs.go#L59-L67). Please advise how to fill gpt structures to match the comment above that example.


aol-nnov:
Also, i'd like to make lint checks more loose in examples, but as failing CI shows, totally disabling them is a bad idea..

deitch:
Looks like vet fails the examples. But I like this setup.

aol-nnov:
I'm out for today... Will try something in near future - tomorrow or may be on Wednesday...

If you have any ideas how to make linter have softer checks for examples, please let me know...

aol-nnov:
@deitch if you like those `check()` and `unused()`, there is [one more action item left](https://github.com/diskfs/go-diskfs/pull/268/files#diff-c2ade2f386c41794d5ebc57ee49b57a5fca8082e03255e5bff13977cbc061287R86) and then it's ready for merge.

deitch:
Yeah, it's nice.

aol-nnov:
Okay, and what about that last action item?

Should `Start` be `1*1024*1024/512` and `Size` be `10*1024*1024`?

To be frank, for me, generally, it's a bit counter-intuitive, that `Start` is in sectors and `Size` is in bytes....

deitch:
> Okay, and what about that last action item?

Oh, I missed that. Sure, those numbers are as good as anything else.

PR #270, opened 2024-12-02
fix: don't match volume label when reading FAT32 directories

### LINKED ISSUE #269: FAT32: path traversal fails if the first directory is equal to the FAT volume label
If the first component of a file path is equal to the FAT32 volume label then the function `readDirWithMkdir` incorrectly matches the volume label entry to the file name being searched for[^1].

A very common scenario where this error can occur is trying to access files in the `/EFI` directory on an ESP partition with volume label `EFI`.

## Steps to reproduce:

- build two FAT32 images (using mtools), one with `LABEL=TEST`, one with `LABEL=EFI`, both containing an `/EFI` directory:

```shell
#!/usr/bin/env bash

set -e

for i in TEST EFI; do
	truncate -s 256M disk_$i
	mformat -i disk_$i -F -v $i ::
	mmd -i disk_$i ::/EFI
	mdir -i disk_$i -/ ::
done
```

- try to read the `/EFI` directory on each using the following

```go
package main

import (
	"fmt"
	"os"

	"github.com/diskfs/go-diskfs/filesystem/fat32"
)

func main() {
	paths := []string{"disk_TEST", "disk_EFI"}

	for _, path := range paths {
		file, err := os.Open(path)
		if err != nil {
			panic(err)
		}
		defer file.Close()

		fs, err := fat32.Read(file, 256*1024*1024, 0, 512)
		if err != nil {
			panic(err)
		}

		dir, err := fs.ReadDir("/EFI")
		if err != nil {
			panic(err)
		}

		fmt.Printf("%#v\n", dir)
	}
}
```

This results in the following output:

```
[]fs.FileInfo{fat32.FileInfo{modTime:time.Time{wall:0x0, ext:63868765252, loc:(*time.Location)(nil)}, mode:0x0, name:"", shortName:".", size:0, isDir:true}, fat32.FileInfo{modTime:time.Time{wall:0x0, ext:63868765252, loc:(*time.Location)(nil)}, mode:0x0, name:"", shortName:"..", size:0, isDir:true}}
panic: error reading directory /EFI: cannot create directory at /EFI since it is a file
```

The expected behaviour would obviously be, that this works for both disk images, regardless of the volume label.

[^1]: https://github.com/diskfs/go-diskfs/blob/15ebb7b3fb4e7106cfad897d1939ccce8f51743a/filesystem/fat32/fat32.go#L997

nkraetzschmar:
fixes #269

### DIFF
--- a/filesystem/fat32/fat32.go
+++ b/filesystem/fat32/fat32.go
@@ -990,6 +990,10 @@ func (fs *FileSystem) readDirWithMkdir(p string, doMake bool) (*Directory, []*di
 		// do we have an entry whose name is the same as this name?
 		found := false
 		for _, e := range entries {
+			// don't match volume label
+			if e.isVolumeLabel {
+				continue
+			}
 			// if the filename does not match, continue
 			// match is determined by any one of:
 			// - long filename == provided name


deitch (approved):
Thanks!

PR #271, opened 2024-12-03
FAT32: add functions to get File cluster chain and disk areas

nkraetzschmar:
For some low-level filesystem operations it can be beneficial to both get the cluster chain of a given file as well as querying where exactly on disk the file contents reside.

Examples where such info might be useful:
- Perform secure erase
- Detect file fragmentation
- Pass Disk byte offset or LBA to a different tool that can work with it

This PR adds two functions, one to just expose the internal getClustersList via the File type.
The other to express the same info directly in disk areas, where each area is described by its starting byte and a length in bytes. (further conversion to disk sectors / LBA can be easily performed by the caller.)

### DIFF
--- a/filesystem/fat32/file.go
+++ b/filesystem/fat32/file.go
@@ -57,7 +57,7 @@ func (fl *File) GetDiskRanges() ([]DiskRange, error) {
 	dataStart := uint64(fs.dataStart)
 
 	var ranges []DiskRange
-	var lastCluster uint32 = 0
+	var lastCluster uint32
 
 	for _, cluster := range clusters {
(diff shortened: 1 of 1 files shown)


deitch (changes requested):
Looks good to me, and I appreciate the description of the use cases. Maybe add those to the docs (even just the inline), so someone looking at this in 6 or 12 months and wonders why we would expose such details, can see?

Also there is a linting error.

nkraetzschmar:
sorry for the delay, added the use cases to the inline docs and fixed the linter errors now

deitch (approved):

deitch:
No problem. Thanks for contributing.

PR #278, opened 2025-01-23
fix: dont include xattr in extended link reading

jonathongardner:
Got an error when trying to read another squashfs file from ubuntu. Not sure if I should have created an issue first. It was raising [this](https://github.com/diskfs/go-diskfs/blob/master/filesystem/squashfs/xattr.go#L23) error. Traced it down to a symlink reading the wrong part for the xattr.

### DIFF
--- a/filesystem/squashfs/inode.go
+++ b/filesystem/squashfs/inode.go
@@ -639,11 +639,12 @@ func parseExtendedSymlink(b []byte) (*extendedSymlink, int, error) {
 	s := &extendedSymlink{
 		links: binary.LittleEndian.Uint32(b[0:4]),
 	}
+	targetSize := int(binary.LittleEndian.Uint32(b[4:8]))
 	// account for the synlink target, plus 4 bytes for the xattr index after it
-	extra = int(binary.LittleEndian.Uint32(b[4:8])) + 4
-	if len(b[target:]) > extra {
-		s.target = string(b[8 : 8+extra])
-		s.xAttrIndex = binary.LittleEndian.Uint32(b[8+extra : 8+extra+4])
+	extra = targetSize + 4
+	if len(b) > extra+target {
+		s.target = string(b[target : target+targetSize])
+		s.xAttrIndex = binary.LittleEndian.Uint32(b[target+targetSize : target+targetSize+4])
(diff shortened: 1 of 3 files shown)


filesystem/squashfs/inode.go
```diff
@@ -639,11 +639,12 @@ func parseExtendedSymlink(b []byte) (*extendedSymlink, int, error) {
 	s := &extendedSymlink{
 		links: binary.LittleEndian.Uint32(b[0:4]),
 	}
+	targetSize := int(binary.LittleEndian.Uint32(b[4:8]))
 	// account for the synlink target, plus 4 bytes for the xattr index after it
-	extra = int(binary.LittleEndian.Uint32(b[4:8])) + 4
-	if len(b[target:]) > extra {
-		s.target = string(b[8 : 8+extra])
```
jonathongardner, line 645: This was reading 4 bytes to much (i.e. it was reading the xattr data as part of the target)

filesystem/squashfs/inode.go
```diff
@@ -639,11 +639,12 @@ func parseExtendedSymlink(b []byte) (*extendedSymlink, int, error) {
...
-		s.xAttrIndex = binary.LittleEndian.Uint32(b[8+extra : 8+extra+4])
```
jonathongardner, line 646: This was reading after the xattr data, so 4 bytes into the next section so it was just a random int

filesystem/squashfs/squashfs.go
```diff
@@ -532,7 +532,7 @@ func (fs *FileSystem) hydrateDirectoryEntries(entries []*directoryEntryRaw) ([]*
 		body, header := in.getBody(), in.getHeader()
 		xattrIndex, has := body.xattrIndex()
 		xattrs := map[string]string{}
-		if has && xattrIndex != noXattrInodeFlag {
+		if has {
```
jonathongardner, line 535: When i was tracing down the bug i noticed all `xattrIndex` already do this check or return false so i removed it

deitch (approved):

deitch:
Quite the catch. Thank you for tracking it down.

PR #280, opened 2025-02-16
Fix panic when tmp path is inccurate

s1gnate-sync:
prior this fix if TMPDIR is set to something like `/tmp//` the iso9660 creation would cause panic.

now it's fixed + test case is added according to @dietch review

### DIFF
--- a/filesystem/iso9660/finalize_test.go
+++ b/filesystem/iso9660/finalize_test.go
@@ -23,6 +23,23 @@ var (
 
 // test creating an iso with el torito boot
 func TestFinalizeElTorito(t *testing.T) {
+	finalizeElTorito(t, "")
+	dir, err := os.MkdirTemp("", "workspace")
+	if err != nil {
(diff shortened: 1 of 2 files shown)


s1gnate-sync:
@deitch had intense sprint and didn't have energy to finish with test-case. Now it's done and I've checked first that test is failing and when added clean to the path and rerun test to ensure it's not failing anymore! All seems good.

There is tiny thing I forgot to add empty line between functions in `finalize_test.go`. Tell me if it's important for you and I will add it ;-)

deitch (approved):

deitch:
Lint doesn't like how you structured the test.

s1gnate-sync:
I've looked how it's handled in other tests and did the same, seems like this rule is useless. I've also added that empty string!

deitch (approved):
Haha, just `nolint`? That is fine. I don't think this is a place where lint is helpful.

s1gnate-sync:
@deitch thanks!

PR #284, opened 2025-03-03
test for squashfs writing and reading errors

deitch:
should incorporate fix as well. Initial test fails, until fix gets in.


deitch:
Nope, it didn't fix that one.

Xynnn007 (approved):

Issue #282, opened 2025-03-03
SquashFS reading is not consistent with writing for large writing


Xynnn007:
When I use the following code to create a SquashFS image and try to write the contents of a file that exceeds the block size, I will find that when reading, I will read the contents of the super block, not just the file contents. If the length of the written data content is less than the block size, there will be no problem

### Example

```go
package main

import (
	"errors"
	"fmt"
	"io"
	"os"

	"github.com/diskfs/go-diskfs"
	"github.com/diskfs/go-diskfs/disk"
	"github.com/diskfs/go-diskfs/filesystem"
	"github.com/diskfs/go-diskfs/filesystem/squashfs"
)

func main() {
	CreateFilesystem := func(d *disk.Disk, spec disk.FilesystemSpec) (filesystem.FileSystem, error) {
		// find out where the partition starts and ends, or if it is the entire disk
		var (
			size, start int64
		)

		switch {
		case spec.Partition == 0:
			size = d.Size
			start = 0
		case d.Table == nil:
			return nil, fmt.Errorf("cannot create filesystem on a partition without a partition table")
		default:
			partitions := d.Table.GetPartitions()
			// API indexes from 1, but slice from 0
			part := spec.Partition - 1
			if spec.Partition > len(partitions) {
				return nil, fmt.Errorf("cannot create filesystem on partition %d greater than maximum partition %d", spec.Partition, len(partitions))
			}
			size = partitions[part].GetSize()
			start = partitions[part].GetStart()
		}

		switch spec.FSType {
		case filesystem.TypeSquashfs:
			return squashfs.Create(d.Backend, size, start, d.LogicalBlocksize)
		default:
			return nil, errors.New("unknown filesystem type requested")
		}
	}

	// Create squashfs image file
	var diskSize int64 = 10 * 1024 * 1024 // 10 MB
	initdataImagePath := "/root/tdx/development/test-squashfs/image-low"
	mydisk, err := diskfs.Create(initdataImagePath, diskSize, 4096)
	if err != nil {
		panic(err)
	}

	fspec := disk.FilesystemSpec{Partition: 0, FSType: filesystem.TypeSquashfs, VolumeLabel: "label"}
	fs, err := CreateFilesystem(mydisk, fspec)
	if err != nil {
		panic(err)
	}

	rw, err := fs.OpenFile("/test", os.O_CREATE|os.O_RDWR)
	if err != nil {
		panic(err)
	}

[...]

PR #286, opened 2025-03-06
squashfs: fix directory inode size

### LINKED ISSUE #285: SquashFS does not set the right filesize for directory inode
Due to SquashFS Spec https://dr-emann.github.io/squashfs/squashfs.html#_directory_inodes Section 5.2, the Directory Inode should have its `file size` 3 bytes larger than the real listing.

SquashFS implementation does not follow this well. Some bugs would come when using `go-diskfs` to generate a squashfs image with a file in it, but reading it using a different tool rather than `go-diskfs` like [backhand](https://github.com/wcampbell0x2a/backhand) in Rust.

Let me put an example for this then.

Xynnn007:
Due to SquashFS spec
https://dr-emann.github.io/squashfs/squashfs.html#_directory_inodes Section 5.2, it says about directory inode file size

'''
	This value is 3 bytes larger than the real listing. The Linux
kernel creates "." and ".." entries for offsets 0 and 1, and only after 3 looks into the listing, subtracting 3 from the size. '''

In go-disks the size is calculated by serialize the 'directory'. A directory serialization includes multiple 'directoryHeader's and 'directoryEntryRaw's. This aligns with the dir_size calculation logic with well-known squasjfs-tools project

https://github.com/plougher/squashfs-tools/blob/master/squashfs-tools/mksquashfs.c#L1549

Note that the file_size is set to 3 larger than the size.

Thus here, we set directoryLocation.size to 3 bytes larger and the new value will be used to set basicDirectory.fileSize.

Fixes #285

### DIFF
--- a/filesystem/squashfs/finalize.go
+++ b/filesystem/squashfs/finalize.go
@@ -1448,7 +1448,7 @@ func populateDirectoryLocations(directories []*finalizeFileInfo) {
 		d.directoryLocation = blockPosition{
 			block:  uint32(pos / int(metadataBlockSize)),
 			offset: uint16(pos % int(metadataBlockSize)),
-			size:   len(b),
+			size:   len(b) + 3,
 		}
 		pos += len(b)
 	}


deitch (approved):
Thanks for fixing.

deitch:
I wonder what the impact of this is on the other issue you found #282. They could be related. Once this is in, I will rebase the test that catches it on this branch and see.

Issue #290, opened 2025-03-27
Error reading Volume Descriptor: mismatched ISO version in Volume Descriptor. Found 2 expected 1


aol-nnov:
Hey, Avi! Long time no see!

Hope this issue finds you well, ha!

I'm facing an issue while trying to run [the example](https://github.com/diskfs/go-diskfs/tree/master/examples/serve-image) with iso images I've created with `genisoimage`, the error is fully stated in the subject.

Could you please tell me, how to diagnose the problem, what extra info could I provide you with?

Let's start with the command

```
genisoimage -f -J -r -volid "xxx_xxxx2_1743003235_src" --publisher "YYYYY YY" --iso-level=4 --joliet-long -o theFile.iso workDir
```

and isoinfo

```
CD-ROM is in ISO 9660 format
System id: LINUX
Volume id: xxx_xxxx2_1743003235_src
Volume set id:
Publisher id: YYYYY YY
Data preparer id:
Application id: GENISOIMAGE ISO 9660/HFS FILESYSTEM CREATOR (C) 1993 E.YOUNGDALE (C) 1997-2006 J.PEARSON/J.SCHILLING (C) 2006-2007 CDRKIT TEAM
Copyright File id:
Abstract File id:
Bibliographic File id:
Volume set size is: 1
Volume set sequence number is: 1
Logical block size is: 2048
Volume size is: 6486
CD-ROM uses ISO 9660:1999 relaxed format

Joliet with UCS level 3 found.
SUSP signatures version 1 found
Rock Ridge signatures version 1 found
Rock Ridge id 'RRIP_1991A'
```

I'm guessing, it's due to `--iso-level=4 --joliet-long`, but I'm unable to change the way those images are created...

Interesting to mention, that if I comment out [the following lines](https://github.com/diskfs/go-diskfs/blob/master/filesystem/iso9660/volume_descriptor.go#L186-L189), everything just works! I've checked the content served and it all seems legit.

I thought, I could implement and contribute joliet support someday, but now I just need a quick local solution...
What are the implications of commenting out that check? Might I lose or unable to read any data from the image?

Thanks in advance!

aol-nnov:
Aha, it's `--iso-level=4` which makes version 2, not `--joliet-long`

man genisoimage says:

>  Level 4 officially does not exist but genisoimage maps it to ISO-9660:1999, which is ISO9660 version 2.
> 
[...]

PR #291, opened 2025-04-05
Support ISO version 2 reading

aol-nnov:
SVDs may have version greater than 1. For now let's check PVD version only.
    
closes https://github.com/diskfs/go-diskfs/issues/290

### DIFF
--- a/filesystem/iso9660/iso9660.go
+++ b/filesystem/iso9660/iso9660.go
@@ -160,7 +160,7 @@ func Read(b backend.Storage, size, start, blocksize int64) (*FileSystem, error) 
 	// we do not do anything with the system area for now
 
 	// next read the volume descriptors, one at a time, until we hit the terminator
-	vds := make([]volumeDescriptor, 2)
+	vds := make([]volumeDescriptor, 0, 128)
 	terminated := false
 	var (
 		pvd *primaryVolumeDescriptor
--- a/filesystem/iso9660/volume_descriptor.go
+++ b/filesystem/iso9660/volume_descriptor.go
@@ -183,8 +183,8 @@ func volumeDescriptorFromBytes(b []byte) (volumeDescriptor, error) {
 	}
 	// validate the version
 	version := b[6]
-	if version != isoVersion {
-		return nil, fmt.Errorf("mismatched ISO version in Volume Descriptor. Found %x expected %x", version, isoVersion)
+	if volumeDescriptorType(b[0]) == volumeDescriptorPrimary && version != isoVersion {
+		return nil, fmt.Errorf("mismatched ISO version in Primary Volume Descriptor. Found %x expected %x", version, isoVersion)
 	}
 	// get the type and data - later we will be more intelligent about this and read actual primary volume info
 	vdType := volumeDescriptorType(b[0])


deitch:
A few questions.

First, on "SVDs may have version greater than 1. For now let's check PVD version only".

Is that the case? PVDs remain locked to 0x01, but SVDs can have a different version? I had not understood that from #290. 

The quote from the manpage is:

```
With  level  4,  an enhanced volume descriptor with version number and file structure version number set to 2 is emitted.  Directory nesting is
not limited to 8 levels, there is no need for a file to contain a dot and the dot has no special meaning, filenames do not  have  version  numbers, and filenames can be up to 207 characters long, or 197 characters if Rock Ridge is used.

When  creating  Version 2 images, genisoimage emits an enhanced volume descriptor, similar but not identical to a primary volume descriptor. Be
careful not to use broken software to make ISO9660 images bootable by assuming a second PVD copy and patching this putative PVD copy into an El
Torito VD.
```

Is it that the PVD remains the same and an SVD is added as enhanced? Or is it that this enhanced VD _replaces_ the PVD? It was not clear to me.

Second, what is [this change](https://github.com/diskfs/go-diskfs/pull/291/files#diff-2a8e67a55539c201c1b00df7ebb69a8cf497e04055fe50b7435db317ddffb98cL163)?

Third, is it possible to add any kind of tests to catch it?

deitch:
Also no idea why lint is failing. But let's get the basics fixed first.

deitch:
I had to bump CI go version, actions versions, lint version, update config, etc. Lots changed. Just rebase yours on latest and lint should be fine.

aol-nnov:
> Is it that the PVD remains the same and an SVD is added as enhanced?

Reading the standard is a quest, yeah. But during my debug session, I've figured out that PVD is always in place and Enhanced VD is placed right after it. This "enhanced vd"/"supplementary vs" is not clear to me, too, but while reading the [wiki page](https://en.wikipedia.org/wiki/ISO_9660), chapter "Volume descriptor set", it seemed to me those terms are equal and have same descriptor type.

So, for ISO9660 version 2, PVD is still exists and has version equals to 1 with SVD descriptor with version 2 next to it. It is still not clear to me if we should totally ignore PVD and rely on SVD in that case. Seems, we need more tests.

> Second, what is [this change](https://github.com/diskfs/go-diskfs/pull/291/files#diff-2a8e67a55539c201c1b00df7ebb69a8cf497e04055fe50b7435db317ddffb98cL163)?

This change is described in the [commit message](https://github.com/diskfs/go-diskfs/pull/291/commits/60c9452ff0d7f9f5f0785435fbd98358275de9c6)

deitch:
OK looks good, thanks for the explanation. Just rebase on latest and push out.

aol-nnov:
Would you like "Fix vds allocation" to be a separate PR?

deitch:
Nope it's fine here as you did it.

deitch:
CI is green. If you are ready, remove from draft?

aol-nnov:
Let it hang for a while.. I decided to make some tests and investigate the interconnection of PVD and enhanced VD in iso v2...

aol-nnov:
Well, the main question for me was if enhanced VD (EVD) is present, should we take path table from PVD or EVD.

I've made a simple test:

```bash
# more than 8 levels of nesting
mkdir -p iso-scratch/1/2/3/4/5/6/7/8/9/10/11/12/13/14/
# file without a dot
echo "oh, hi!" > iso-scratch/1/2/3/4/5/6/7/8/9/10/11/12/13/14/file

genisoimage --iso-level=4 -o v2.iso iso-scratch/
```

For testing purposes I've replaced the [following lines](https://github.com/diskfs/go-diskfs/blob/master/filesystem/iso9660/iso9660.go#L202-L214) with a snippet below:

```go
	evd, _ := vds[1].(*supplementaryVolumeDescriptor)

	rootDirEntry = evd.rootDirectoryEntry
	pathTableBytes := make([]byte, evd.pathTableSize)
	pathTableLocation := evd.pathTableLLocation * uint32(evd.blocksize)
	read, err = b.ReadAt(pathTableBytes, int64(pathTableLocation))
	if err != nil {
		return nil, fmt.Errorf("evd: unable to read path table of size %d at location %d: %v", evd.pathTableSize, pathTableLocation, err)
	}
	if read != len(pathTableBytes) {
		return nil, fmt.Errorf("evd: read %d bytes of path table instead of expected %d at location %d", read, pvd.pathTableSize, pathTableLocation)
	}
	pt = parsePathTable(pathTableBytes)
```

It gets path table from EVD instead of PVD.
This test showed no difference if we take path table from PVD or EVD, yet, it did not answer to me which approach is correct )))

So let's call it a day, merge it as is and see if it broke someday.
Any objections, Avi?

deitch (approved):

deitch:
Yes, looks good to me. Thank you.

aol-nnov:
Nice! Would you mind making a point release with this improvement?

deitch:
Sure done.

PR #297, opened 2025-05-23
feat(fat32): Add file system attribute setting and checking functions

xspeed1989:
Added the ability to set and check file attributes for the FAT32 file system, including system files, hidden files, and read-only files. These attributes are correctly encoded and decoded in directory entries to ensure persistence and consistency of file attributes.


deitch (dismissed):

deitch:
Thanks.

deitch:
Some lint errors. What is the `on bool` for in those functions?

xspeed1989:
> Some lint errors. What is the `on bool` for in those functions?

There are some getter functions, I will remove the unnecessary param

deitch (approved):

deitch:
Solid, thanks.

Issue #299, opened 2025-06-23
squash fs return wrong data for a file


jonathongardner:
Reading date from a file doesn't match what is expected. I originally found this trying to read some squashfs file in ubuntu. I was able to copy the file out into a single squashfs file and still see the issue. Here is the code:
```golang
package main

import (
	"crypto/sha1"
	"encoding/hex"
	"fmt"
	"io"
	"os"

	"github.com/diskfs/go-diskfs"
	dskfile "github.com/diskfs/go-diskfs/backend/file"
)

func main() {
	if len(os.Args) != 3 {
		panic("Usage: go run main.go <squashfs_file> [starting-path]")
	}

	printDiskHash(os.Args[1], os.Args[2])
}

func printDiskHash(file, toHashFile string) {
	//--------------Open the disk-----------------
	src, err := os.Open(file)
	if err != nil {
		panic(fmt.Errorf("error opening source file %v", err))
	}
	defer func() {
		if tmpErr := src.Close(); tmpErr != nil && err == nil {
			panic(fmt.Errorf("error closing source file: %v", tmpErr))
		}
	}()

	disk, err := diskfs.OpenBackend(dskfile.New(src, true), diskfs.WithSectorSize(4096))
	if err != nil {
		panic(fmt.Errorf("error getting backend %v", err))
	}

	if disk.Table != nil {
		panic("Disk table is not nil, expected a single filesystem without partitions")
	}

	fs, err := disk.GetFilesystem(0)
	if err != nil {
		panic(fmt.Errorf("error getting filesystem %v", err))
	}

	//--------------Open file in disk-----------------
	f, err := fs.OpenFile(toHashFile, os.O_RDONLY)
	if err != nil {
		panic(fmt.Errorf("error opening disk file %s: %v", toHashFile, err))
	}
	defer func() {
		if tmpErr := f.Close(); tmpErr != nil {
			panic(fmt.Errorf("error closing disk file %s: %v", toHashFile, tmpErr))
		}
	}()

	//--------------Calculate disk hash-----------------
	hasher := sha1.New()

	v, err := io.Copy(hasher, f)
	if err != nil {
		panic(fmt.Errorf("error getting disk file hash %s: %v", toHashFile, err))
	}

	sum := hasher.Sum(nil)

	fmt.Printf(
		"file hash: %s, size: %d \n",
		hex.EncodeToString(sum),
		v,
	)
}
```

```sh
go run main.go 

[...]

PR #300, opened 2025-06-24
fix: handle caching sparse reading block

jonathongardner:
Fix: #299

Cache the previous block size too, so when deciding to use the cached block can compare size to handle sparse blocks


deitch:
@cusspvz is this good to go?

deitch:
I am fine with it.

PR #315, opened 2025-12-11
iso9660: Truncate short filenames to 8.3 format for compatibility

zaneb:
When creating ISO 9660 filesystems with Rock Ridge extensions, diskfs creates both a Rock Ridge long name and an ISO 9660 short name for compatibility with tools that don't support Rock Ridge.

Previously, the calculateShortnameExtension() function would uppercase and sanitize filenames but did not enforce the ISO 9660 Level 1 8.3 format constraints. This resulted in short names with extensions longer than 3 characters and basenames longer than 8 characters.

Some tools access files using their ISO 9660 short names and expect proper 8.3 format (8-character basename, 3-character extension). When diskfs creates ISOs with longer extensions or basenames, these tools fail to find the files.

This commit updates calculateShortnameExtension() to enforce ISO 9660 Level 1 limits:
- Extensions are truncated to a maximum of 3 characters
- Basenames are truncated to a maximum of 8 characters

This matches the behavior of traditional ISO creation tools like xorriso (default Level 1 mode) and ensures maximum compatibility.

Comprehensive tests have been added to verify the 8.3 format is correctly enforced for various filename patterns including long names, special characters, and edge cases.

Assisted-by: Claude Code

### DIFF
--- a/filesystem/iso9660/finalize.go
+++ b/filesystem/iso9660/finalize.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"math"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


deitch (dismissed):

zaneb:
OK, I added collision prevention, so according to my testing this should do the same thing as xorriso in Level 1 mode now.

deitch (dismissed):

PR #317, opened 2025-12-19
feat: support reproducible images

frezbo:
Support reproducible fat32 images as like `mkfs.vfat --invariant`.

This also supports creating files and directories with with the value of `SOURCE_DATE_EPOCH` when set as per https://reproducible-builds.org/ providing a way to create reproducible filesystems.

### DIFF
--- a/filesystem/fat32/fat32_test.go
+++ b/filesystem/fat32/fat32_test.go
@@ -8,14 +8,17 @@ package fat32_test
 import (
 	"bytes"
 	"crypto/rand"
+	"crypto/sha256"
 	"fmt"
 	"io"
 	mathrandv2 "math/rand/v2"
 	"os"
 	"path"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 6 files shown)


deitch:
Also, lint failed.

filesystem/fat32/fat32_test.go
```diff
@@ -102,15 +105,6 @@ func tmpFat32(fill bool, embedPre, embedPost int64) (*os.File, error) {
 	return f, nil
 }
 
-func TestFat32Type(t *testing.T) {
```
frezbo: this was a rebase error, will fix

deitch:
Oh, I think I see. You are not worrying about the timestamp when you create the filesystem (although that matters, too), but (also) about what happens when I already have a filesystem (accessed via `Create` where you created a new one, or via `Get` where you access an already existing one), and you want new directory or file creation to control the timestamp, rather than using `time.Now()` to use the actual current clock time. Is that it?

PR #318, opened 2025-12-22
fix: buffered writes when using `io.Copy`

frezbo:
When copying files to a fat filesystem and if using `io.Copy` on the `filesystem.OpenFile` it was slow which the entries were flushed for each write which is in-efficient. Only flush once on `Close()`.

### DIFF
--- a/filesystem/fat32/fat32.go
+++ b/filesystem/fat32/fat32.go
@@ -60,6 +60,7 @@ type FileSystem struct {
 	size            int64
 	start           int64
 	backend         backend.Storage
+	fatDirty        bool // tracks if FAT needs to be flushed
 }
 
 // Equal compare if two filesystems are equal
@@ -351,7 +352,6 @@ func Read(b backend.Storage, size, start, blocksize int64) (*FileSystem, error) 
 		return nil, fmt.Errorf("only could read %d bytes from file", n)
 	}
 	bs, err := msDosBootSectorFromBytes(bsb)
-
 	if err != nil {
 		return nil, fmt.Errorf("error reading MS-DOS Boot Sector: %w", err)
 	}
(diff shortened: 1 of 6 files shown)


deitch:
OK, so the problem is that `Write(b []byte)` treats its argument as complete, which is the purpose of write. As such, it thinks it knows the entire size to write (`len(b)`), allocates the clusters for that size, and is done. But if you actually have a total size of `1000*len(b)`, then the write itself is fine, but it will allocate the clusters 1000 times, which is slow and inefficient. If it only knew, "this byte slice I am getting is not the whole thing, but one of many," then it might be more efficient about it.

You are trying to fix this by having it implement `ReadFrom()`, using it as a way to pass it an `io.Reader`, hence a stream, and use that to say, "this might not be everything you are going to read, I gave you a stream instead of a slice, so you know that." And then inside `ReadFrom()`, you add logic to try and guess the total size if you can; if not, then use an allocation algorithm that tries to be efficient about it.

Now that I get it (although I haven't dug into your allocation algorithm), I have a few questions.

First, if `src` implements `WriterTo`, then all of this will not work, since `io.Copy(dst, src)` calls `src.WriteTo()` before even trying `dst.ReadFrom()`. So even if this logic works, it needs to be implemented equally for `Write()`. Which does not obviate the need for `ReadFrom()` - it is a good thing to do - but it does mean the heuristic needs to be used for both `Write()` and `ReadFrom()`. I get that it is easier with `ReadFrom()` - single call, you know when you are done, can preallocate at the beginning and cleanup at the end - but we would need to account for both cases.

Second, do we know algorithm the fat32 driver uses, maybe in Linux? It must have a similar issue when someone calls [write(2)](https://man7.org/linux/man-pages/man2/write.2.html), which eventually sends data to a file descriptor, which is held open by the kernel for a particular filesystem driver, i.e. fat32.

A naive approach might be similar to what you did: always create a large number of clusters, keep them cached, use them from the cache, when the cache runs dry, allocate another large chunk. I don't recall offhand how fat32 filesystem handles clusters created but not for a particular file, if that is doable.

PR #352, opened 2026-03-26
ext4: implement extended attribute reading

vvoland:
- needs: https://github.com/diskfs/go-diskfs/pull/351

Add GetXattr(path) public method that reads both inline (ibody) and external block xattrs.

Support all standard name index prefixes (user, trusted, security, system, posix_acl).


vvoland:
Still relevant

deitch (approved):
I appreciate the additional tests especially.

PR #357, opened 2026-03-28
fix(iso9660): Open() now handles directories and implements fs.ReadDirFile

### LINKED ISSUE #356: iso9660: `Open()` rejects directories, making `http.FileServer(http.FS(fs))` unable to browse directories
### What happened

In v1.8.0, `iso9660.FileSystem` now implements `fs.FS` via the `Open()` method. However, `Open()` delegates to `OpenFile()`, which explicitly rejects directories:

```go
// iso9660.go line 443-444
if dir == filename {
    return nil, fmt.Errorf("cannot open directory %s as file", p)
}
```

This means calling `Open("boot")` on a directory returns an error instead of an `fs.File` that implements `fs.ReadDirFile`.

### Impact

Go's `http.FileServer(http.FS(fs))` relies on `Open()` to handle both files and directories. When a user requests a directory URL (e.g., `/boot/`), `http.FileServer` calls `fs.Open("boot")`, gets an error, and returns HTTP 500.

This makes the official [serve-image example](https://github.com/diskfs/go-diskfs/blob/master/examples/serve-image/main.go) unable to serve directory listings for iso9660 filesystems.

### How to reproduce

```go
package main

import (
    "fmt"
    "github.com/diskfs/go-diskfs"
    "github.com/diskfs/go-diskfs/backend/file"
    "github.com/diskfs/go-diskfs/filesystem/iso9660"
    "os"
)

func main() {
    f, _ := os.Open("test.iso")
    defer f.Close()
    fi, _ := f.Stat()
    b := file.New(f, true)
    fs, _ := iso9660.Read(b, fi.Size(), 0, 0)

    // This fails with: "cannot open directory boot as file"
    _, err := fs.Open("boot")
    fmt.Println(err)
}
```

### Expected behavior

`Open()` on a directory path should return an `fs.File` that also implements `fs.ReadDirFile`, allowing directory listing via `ReadDir(n int)`.

### What I think the fix could look like

`Open()` (or `OpenFile()`) should detect directory paths and return a directory file handle, similar to what other filesystem implementations (fat32, ext4) do. A minimal directory handle would need to:

1. Implement `fs.File` (Stat, Read, Close)
2. Implement `fs.ReadDirFile` (ReadDir with cursor/offset tracking)
3. `Read()` should return `fs.ErrInvalid` (directories are not readable as byte streams)

### Environment

- go-diskfs version: v1.8.0
- Go version: 1.25.7
- OS: macOS / Linux

radioinmyhead:
Fixes #356

### Changes:
1. **New file: `filesystem/iso9660/dirfile.go`**
   - `dirFile` struct — implements both `fs.File` and `fs.ReadDirFile` for directory handles.
   - `Read()` returns `fs.ErrInvalid` (directories aren't byte-readable).
   - `Close()` marks the handle as closed.
   - `Stat()` returns the directory entry's FileInfo; for root `.`, wraps with `rootDirInfo` to fix `Name()`.
   - `ReadDir(n)` — lazy-loads entries via `FileSystem.ReadDir()` and tracks a cursor for sequential calls, following the full `fs.ReadDirFile` contract.
   - `rootDirInfo` wrapper — overrides `Name()` to return `.` (the raw root `directoryEntry` has empty filename).

2. **Modified: `filesystem/iso9660/iso9660.go`**
   - `Open()` rewritten — no longer delegates to `OpenFile()`. Instead:
     - In workspace mode: calls `os.Open()` directly.
     - For `.`: returns `dirFile` wrapping `rootDir`.
     - For other paths: reads the parent directory entries, and returns either a `dirFile` (if directory) or a `File` (if regular file).
   - `OpenFile()` is unchanged — existing internal API preserved.
   - `Stat()` fixed — added special case for `.` to return root directory info.

3. **Modified: `filesystem/iso9660/compatibility_test.go`**
   - Added tests: `TestOpenRootDirectory`, `TestOpenSubdirectory`, `TestStatRootDirectory`, `TestOpenDirectoryRockRidge`.

### Test results
- [x] All existing tests pass
- [x] All new tests pass
- [x] Full project builds clean

### DIFF
--- a/filesystem/iso9660/compatibility_test.go
+++ b/filesystem/iso9660/compatibility_test.go
@@ -1,6 +1,7 @@
 package iso9660
 
 import (
+	"io/fs"
 	"os"
 	"testing"
 
@@ -16,25 +17,25 @@ func TestFSCompatibility(t *testing.T) {
 	defer f.Close()
 
 	b := file.New(f, true)
-	fs, err := Read(b, 0, 0, 2048)
+	fsys, err := Read(b, 0, 0, 2048)
 	if err != nil {
 		t.Fatalf("iso read: %s", err)
 	}
 
-	if _, err := fs.ReadDir("/"); err == nil {
+	if _, err := fsys.ReadDir("/"); err == nil {
 		t.Fatalf("should have given error with ReadDir(/): %s", err)
 	}
-	entries, err := fs.ReadDir(".")
+	entries, err := fsys.ReadDir(".")
 	if err != nil {
 		t.Fatalf("should not have given error with ReadDir(.): %s", err)
 	}
 	if len(entries) != 5 {
 		t.Fatalf("should be 5 entries in iso fs")
 	}
-	if _, err := fs.Open("/README.MD"); err == nil {
+	if _, err := fsys.Open("/README.MD"); err == nil {
 		t.Fatalf("should have given an error with Open(/README.MD)")
 	}
-	testfile, err := fs.Open("README.MD")
+	testfile, err := fsys.Open("README.MD")
 	if err != nil {
 		t.Fatalf("test file: %s", err)
 	}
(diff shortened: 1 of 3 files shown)


deitch (approved):
Took me a bit, but I get it. Thanks

Issue #293, opened 2025-04-21
FAT12 and FAT16 support?


acheronfail:
Is supporting FAT12/16 something this project is interested in doing?

I have a use-case for mounting some disks which have a some FAT12 and FAT16 partitions, with one larger FAT32 partition.
If I remember correctly, FAT32 is an extension of FAT12/16, so I presume that many (if not all) of the building blocks necessary for supporting FAT12/16 should already be here?

deitch:
Yes, that would be a good addition. You are correct that one FAT is an extension of the other. It is documented pretty decently [in Wikipedia](https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system), linked from [filesystem/fat32/doc.go](https://github.com/diskfs/go-diskfs/blob/master/filesystem/fat32/doc.go). FAT32 Extended BPB includes and extends FAT12/16 Extended BPB, which extends DOS 3.31 BPB, which extends DOS 3.2 BPB, which extends DOS 3.0 BPB. The FSIS is new to FAT32, but a lot of the structures are similar. The reserved sectors and FAT tables can be different in size, if I recall correctly, but that is manageable. If you look at [the files in this library implementing fat32](https://github.com/diskfs/go-diskfs/tree/master/filesystem/fat32), you can see that each substructure exists. 

So it should be doable without too much pain.

Do you want to give it a shot?

acheronfail:
@deitch in terms of implementation - how do you feel we should organise the code? Since the `fat32` package already has a lot of the building blocks that could be shared between the other FAT types, should we move a lot of the Boot sector/BPB stuff into a common `fat` package which can be shared between `fat32` (and the new `fat12`, etc packages)?

deitch:
I have been thinking about it. I don't think it all should be in a single package (directory), as it is less clean than the current structure, which has each filesystem type in its own. Also, when you [`Disk.GetFilesystem()`](https://github.com/diskfs/go-diskfs/blob/master/disk/disk.go#L218-L244), it tries each directory, e.g. `fat32.Read()` and `iso9660.Read()`, etc. If that is hard, we can change that, e.g. `fat32.Read(fat32)` or `fat32.Read(fat16)`, etc. Or we can have `fat32.Read()` have its own internal analyser that figures out, "oh this FAT is fat12" or "this FAT is fat32".

I would not mind moving each one into its own directory without the common `fat` directory:

```
|- filesystem
    |- fat12
    |- fat16
    |- fat32
```

[...]

PR #358, opened 2026-03-30
filesystem: add FAT12 and FAT16 support

waldner:
Adds full read/write support for FAT12 and FAT16 filesystems, structured
as three separate packages (`fat12`, `fat16`, `fat32`) where each builds on the
previous. The existing `fat32` package is refactored to eliminate duplication
by embedding the new `fat12` base.

This is the implementation discussed in issue #293.

---

## Architecture

```
filesystem/fat12  — base package containing all shared logic
filesystem/fat16  — thin package (~250 lines) that embeds fat12.FileSystem
filesystem/fat32  — refactored to embed fat12.FileSystem; retains only
                    FAT32-specific code (dos71EBPB, FSIS, backup sectors)
```

The three FAT variants differ in four concrete ways that are handled without duplication:

1. **FAT table entry width** — 12-bit packed / 16-bit / 32-bit. Abstracted behind a `fat12.FATTable` interface; each variant provides its own implementation. `fat12` also exports `NewFat16Table()` so the `fat16` package does not need to duplicate the 16-bit logic.

2. **Root directory storage** — FAT12/16 use a fixed-size flat region between the FAT copies and the data area; FAT32 uses a regular cluster chain. `fat12.FileSystem` holds `rootDirOffset` and `rootDirMaxEntries` fields; when the latter is zero the cluster-chain path is taken (FAT32).

3. **Boot parameter block layout** — FAT12/16 use the DOS 4.0 EBPB (the new `fat12.Dos40EBPB` type); FAT32 keeps its own `dos71EBPB`. `fat12` exports the shared BPB layers (`Dos20BPB`, `Dos331BPB`) that `fat32` previously duplicated locally.

4. **FAT32-specific structures** — FSInformationSector, backup boot sector. Integrated via two function-field hooks on `fat12.FileSystem`:
   - `WriteBootSectorFn` — replaces the default boot-sector write; `fat32` uses this to write the `dos71EBPB` format plus the backup at sector 6.
   - `AfterWriteFAT` — called after every FAT flush; `fat32` uses this to keep the FSInformationSector in sync.

---

## New: `filesystem/fat12`

**Exported shared types** (imported by `fat16` and `fat32`):
- `Dos20BPB`, `Dos331BPB` — the DOS 2.0 and 3.31 BPBs shared by all variants.
- `Dos40EBPB` — the DOS 4.0 Extended BPB used by FAT12 and FAT16.
- `FATTable` — interface abstracting on-disk FAT entry encoding.
- `FileSystem` — the full filesystem implementation.
- `NewFileSystem()` — exported constructor; `fat16` and `fat32` inject their own `FATTable` and hooks to reuse all methods.
- `NewFat16Table()` — constructs a 16-bit FATTable for use by `fat16`.
- `SetRootDirLabel()` — split from `SetLabel` so `fat32` can reuse the root-directory label-entry update logic without duplicating it.
- `Backend()`, `Start()`, `BytesPerCluster()`, `DataStart()` — accessors for `fat32`'s write helpers.

**Cluster-size table** (keeps cluster count < 4085 per the Microsoft FAT spec):

| Volume size | Sectors/cluster |
|---|---|
| ≤ 2 MB | 1 |
| ≤ 4 MB | 2 |
| < 8 MB | 4 |
| ≤ 16 MB | 8 |
| ≤ 32 MB | 16 |
| ≤ 64 MB | 32 |
| > 64 MB | 64 |

Note: the strict `< 8 MB` boundary for s/c=4 is intentional — exactly 8 MB with 4 sectors/cluster produces 4086 clusters, one above the FAT12 limit of 4084.

`Read()` self-validates by computing the data cluster count from the BPB and returning an error when count ≥ 4085, allowing `disk.GetFilesystem()` to fall through to `fat16` or `fat32` cleanly.

**TODOs fixed (inherited from `fat32`):**

*Short-name numeric-tail uniqueness (`createEntry`):*
Previously `convertLfnSfn` always emitted `~1` when a long filename stem exceeded 8 characters. Two files whose names truncated to the same 6-char stem would silently produce duplicate short names, corrupting the directory. A new `uniqueShortName()` function implements the FAT spec numeric-tail algorithm:
- counter 1–9 → `stem[:6]+"~N"`
- counter 10–99 → `stem[:5]+"~NN"`
- counter 100–999 → `stem[:4]+"~NNN"`
- …and so on

*Long/short-name matching in `removeEntry` / `renameEntry`:*
Both functions previously matched only on `filenameLong`. A new `matchesFilename()` helper checks both the long filename and the reconstructed 8.3 short name (case-insensitively), so `Remove()` and `Rename()` work correctly when the caller supplies the short-name form. `renameEntry` also calls `uniqueShortName` for the new filename, excluding the old entry from the conflict scan so its outgoing slot is treated as available.

---

## New: `filesystem/fat16`

`FileSystem` embeds `*fat12.FileSystem` and overrides only `Type()`. All filesystem operations (`OpenFile`, `ReadDir`, `Mkdir`, `Remove`, `Rename`, `Label`, `SetLabel`, `allocateSpace`, …) are inherited with no duplication.

**Cluster-size table** (keeps cluster count in [4085, 65524]):

| Volume size | Sectors/cluster |
|---|---|
| ≤ 32 MB | 2 |
| ≤ 128 MB | 4 |
| ≤ 256 MB | 8 |
| ≤ 512 MB | 16 |
| ≤ 1 GB | 32 |
| > 1 GB | 64 |

`Read()` self-validates that 4085 ≤ clusters < 65525, returning an error otherwise.

---

## Changed: `filesystem/fat32`

`FileSystem` now embeds `*fat12.FileSystem`:

```go
type FileSystem struct {
    *fat12.FileSystem
    fsis     FSInformationSector
    bpbFat32 *dos71EBPB
}
```

**Deleted** (now inherited from `fat12.FileSystem`):
- `directoryentry.go` (503 lines) — byte-for-byte identical to `fat12`'s.
- `file.go` (319 lines) — byte-for-byte identical.
- `fileinfo.go` (88 lines) — byte-for-byte identical.
- `directory.go` (154 lines) — subset of `fat12`'s.
- All 30 high-level `FileSystem` methods from `fat32.go` (~800 lines): `Close`, `Mkdir`, `Mknod`, `Link`, `Symlink`, `Chmod`, `Chown`, `Chtimes`, `ReadDir`, `Open`, `OpenFile`, `ReadFile`, `Remove`, `Rename`, `Stat`, `Label`, `getClusterList`, `readDirectory`, `writeDirectoryEntries`, `mkSubdir`, `mkFile`, `mkLabel`, `readDirWithMkdir`, `allocateSpace`.

**Deleted** (now shared via `fat12` exports):
- `dos20bpb.go`, `dos331bpb.go` — replaced by imports of `fat12.Dos20BPB` and `fat12.Dos331BPB`. `dos71bpb.go` updated to use these types. Test files rewritten accordingly.

**Deleted** (tested by `fat12`'s own test suite):
- `fat32_internal_test.go` (455 lines), `directoryentry_internal_test.go`, `directory_internal_test.go`, `fileinfo_internal_test.go`.

**Remaining in `fat32`** (genuinely FAT32-specific):
`dos71EBPB` + BPB parsing, `FSInformationSector`, `msDosBootSector` (FAT32 variant), 32-bit FAT table, `Create()` / `Read()`, `writeBootSector()` / `writeFsis()`, `SetLabel()` / `Type()` overrides.

`SectorSize` is redeclared as a type alias (`= fat12.SectorSize`) so that `fat32.SectorSize512` remains usable in `fat12.Dos20BPB` struct literals without an explicit conversion, preserving the exported constant for callers.

`fat32.table` now implements `fat12.FATTable` (`ClusterValue`, `SetCluster`, `IsEOC`, `EOCMarker`, `UnusedMarker`, `MaxCluster`, `FATID`, `RootDirCluster`, `Size`, `FromBytes`, `Bytes`). A compile-time interface guard is added.

---

## Changed: `filesystem/filesystem.go`

Added `TypeFat12` and `TypeFat16` at the end of the `Type` const block. `TypeFat32`'s iota value (0) is unchanged; no existing callers are affected.

---

## Changed: `disk/disk.go`

`CreateFilesystem()` dispatches `TypeFat12` and `TypeFat16` to the new packages.

`GetFilesystem()` now attempts fat32 → fat16 → fat12 in order. Each `Read()` self-validates its cluster-count range and returns a typed error on mismatch, so the fall-through is clean without shared detection logic.

---

### `filesystem/fat12/table_internal_test.go` (white-box, 15 tests)
12-bit packing round-trips including the three-byte-sharing invariant; EOC detection for all boundary values; `fat12Table` and `fat16TableImpl` `Bytes()`/`FromBytes()` round-trips. Found and fixed an off-by-one in `fat16TableImpl.Bytes()` (`i <= max` instead of `i < max` → slice-bounds panic).

### `filesystem/fat12/dos40ebpb_internal_test.go` (white-box, 10 tests)
`Dos40EBPB` parse/serialise round-trips for both short (0x28) and long (0x29) forms, error cases, `TotalSectors()` field-preference logic. `Dos20BPB` and `Dos331BPB` parse/round-trip tests.

### `filesystem/fat12/directory_internal_test.go` (white-box, 20 tests)
`uniqueShortName()` — no conflicts, counter progression through ~1/~9/~10/~100, volume-label exclusion, short stems. `matchesFilename()` — long name exact/case-insensitive, short name exact/case-insensitive, no-LFN entries, non-matching names. `createEntry` — two colliding long names get distinct tails, 12 consecutive collisions all distinct, non-truncated name has no tail. `removeEntry` / `renameEntry` — by long name, by short name, new-name collision resolved, old slot freed for reuse, existing destination overwritten, not-found errors.

### `filesystem/fat12/fat12_test.go` (black-box, 20 tests)
`Create` (valid sizes + expected errors), `Type`, `Read` (round-trip and rejection of FAT16 images), `Label`/`SetLabel`, `Mkdir`, single-cluster and multi-cluster file write/read, subdirectory file write/read, `ReadDir`, `Remove`, `Rename`, `Stat`, `ReadFile`, `Truncate`, root-directory overflow, reproducible build, `fs.FS` interface including `iofs.WalkDir`. Found and fixed incorrect FAT12 cluster-size boundaries.

### `filesystem/fat16/fat16_test.go` (black-box, 22 tests)
Same coverage as FAT12 tests plus: `TypeNotFat12`, 64 KB multi-cluster file, 3-level deep subdirectory, `ManyFiles` (5 dirs × 10 files verified after reopen). `Create` tests cover 32 MB, 128 MB and 512 MB images. Found and fixed incorrect FAT16 cluster-size boundaries.


deitch:
Great. I am really happy to see this. I have a few comments which I will put inline, but let's get CI to run anyways.

waldner:
> @deitch @waldner Just to play the devil's advocate here, and please excuse me for making assumptions, but both the code and PR description seem heavily LLM generated (Claude Code?). How much has this been tested in practice?

They are, and I was honest about this upfront (https://github.com/diskfs/go-diskfs/issues/293#issuecomment-4150139577). If you see the commit description, existing tests still pass, and many new tests have been added. I've been using the modified code to generate FAT12 ESP partitions embedded in ISO images and so far have not had any issue.

PR #366, opened 2026-04-02
ext4: add benchmarks mimicking creating a containerd rwlayer

djs55:
The first patch is #365 (to fix a journal issue). The second patch is new.

Thinking of nerdbox + containerd on Mac, where the writable layer is an ext4 filesystem created on the host: currently it uses `mkfs.ext4`, but it would be interesting if it could use `go-diskfs` instead.

This PR adds a benchmark, comparing it to `mkfs.ext4`:

```
go-diskfs % PATH="/opt/homebrew/opt/e2fsprogs/sbin:$PATH" \
        go test -bench BenchmarkCreate20GB -run '^$' -benchmem ./filesystem/ext4/
goos: darwin
goarch: arm64
pkg: github.com/diskfs/go-diskfs/filesystem/ext4
cpu: Apple M3 Max
BenchmarkCreate20GB-14                         2         515633521 ns/op        763341676 B/op   2789173 allocs/op
BenchmarkCreate20GBNoJournal-14                3         447836097 ns/op        762175229 B/op   2788775 allocs/op
BenchmarkCreate20GBMkfsExec-14               115          10447870 ns/op            9808 B/op         58 allocs/op
PASS
ok      github.com/diskfs/go-diskfs/filesystem/ext4     3.982s
```

(I added the unit test just to check the outputs are valid, separate from the benchmark itself)

So it's a bit slower (see the ns/op) but I've got some experimental patches which could speed it up. I'll make those as a separate draft PR and describe how I created them, in case you'd like to reproduce (since it was an AI investigation).

### DIFF
--- a/filesystem/ext4/create_bench_test.go
+++ b/filesystem/ext4/create_bench_test.go
@@ -1,0 +1,188 @@
+package ext4
+
+import (
+	"bytes"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+
+	"github.com/diskfs/go-diskfs/backend/file"
+)
+
+func checkE2fsck(tb testing.TB) {
+	tb.Helper()
+	if _, err := exec.LookPath("e2fsck"); err != nil {
+		tb.Skip("e2fsck not available")
+	}
+}
+
+func checkMkfsExt4(tb testing.TB) {
+	tb.Helper()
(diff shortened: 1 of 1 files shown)


deitch:
Needs a rebase, and lint is failing

Issue #73, opened 2020-11-04
Feature Request: Joliet and Rock Ridge extensions supporting


vdmvoro:
Good day
Hope you are doing well.

For now: Joliet and Rock Ridge extensions are not supported (I mean reading or writing).
As I see you have a plans:

- Joliet extensions to iso9660
- Rock Ridge sparse file support - supports the flag, but not yet reading or writing

Do you plan to write additional methods for this in near future? Maybe you know some ETA?

deitch:
At some point, yes. squashfs writing and ext4 writing have been bogged down for a while. I would like to see Joliet and RR in as well. 

Would be more than happy to take contributions/Pull Requests. If not, it is just a question of finding time.

davidnewhall:
TL;DR: I found 1 [old library](https://github.com/dgodd/iso9660/tree/master) that has Rock Ridge support, and 1 that [supports UDF](https://github.com/mogaika/udf). I found none that support Joliet.

Recently I did a lot of googling and GitHub searching to find libraries that allow me to extract ISO images. The real problem is most of my users download random stuff and the image types are all over the place. I quickly found that there aren't many libraries out there, and none of them support _everything_ yet. Maybe one day we'll get there. I just wanted to share some of my research in case anyone finds time to write the implementation(s). I'd put more time into it, but I'm far from an expert on file systems, and a lot of this is very new to me. I work best from pointers and examples, so if there's anything to add, please throw it in here.

The info I found is spread through [a couple comments on this issue](https://github.com/Unpackerr/unpackerr/issues/264#issuecomment-1527016993).

deitch:
I could point you at all of the structures for writing disk images and filesystems here, but RR and Joliet are extensions to an existing filesystem format, not a new one. The references all are linked in [this brief doc](https://github.com/diskfs/go-diskfs/blob/master/filesystem/iso9660/doc.go)

luthermonson:
@deitch this can be closed

PR #372, opened 2026-04-10
Joliet Support

luthermonson:
Architecture Document: https://gist.github.com/luthermonson/387fbca33389f947c99d86df930df5dd

 ## Summary

- Add Joliet extension support for ISO 9660, enabling long Unicode filenames (up to 64 characters) via a Supplementary Volume Descriptor (SVD) with UCS-2 encoding
- On read, automatically uses the Joliet directory tree when present and Rock Ridge is not active, preserving original filenames with mixed case, spaces, and long names
- On write, `FinalizeOptions{Joliet: true}` generates a complete Joliet SVD with its own directory tree and path tables, sharing file data extents with the PVD

## Details

**Write path (`finalize.go`):**
- New `Joliet` flag on `FinalizeOptions`
- Builds a parallel Joliet directory tree with UCS-2 encoded filenames
- Generates Joliet-specific L/M path tables
- Writes a Supplementary Volume Descriptor with UCS-2 Level 3 escape sequences

**Read path (`iso9660.go`):**
- Detects Joliet SVDs by checking for UCS-2 escape sequences (`%/E`, `%/C`, `%/@`)
- Loads Joliet path table and root directory entry
- `readDirectory` transparently dispatches to Joliet tree when available (and Rock Ridge is not active)

**Supporting changes:**
- `directoryEntry` gains a `joliet` flag for UCS-2 filename encoding/decoding
- `pathTable` gains Joliet-specific serialization and parsing (`toJolietLBytes`/`toJolietMBytes`, `parseJolietPathTable`)
- `volumeDescriptors` tracks the supplementary descriptor; SVD `toBytes`/`parseSupplementaryVolumeDescriptor` now handle escape sequences, volume flags, and file structure version

## Test plan

- [x] Unit test: `TestFinalizeJolietRoundTrip` ΓÇö creates an ISO with Joliet, reads it back via go-diskfs, verifies long/mixed-case filenames, subdirectories, and file content
- [x] Integration test: `joliet_xorriso_test.go` ΓÇö builds ISOs with xorriso, reads them with go-diskfs; also builds ISOs with go-diskfs and verifies them with xorriso, covering long filenames, Unicode names, deeply nested paths, and mixed Joliet+Rock Ridge scenarios

### DIFF
--- a/filesystem/iso9660/directoryentry.go
+++ b/filesystem/iso9660/directoryentry.go
@@ -38,6 +38,7 @@ type directoryEntry struct {
 	hasMoreEntries           bool
 	isSelf                   bool
 	isParent                 bool
+	joliet                   bool
 	volumeSequence           uint16
 	filesystem               *FileSystem
 	filename                 string
@@ -53,7 +54,11 @@ func (de *directoryEntry) countNamelenBytes() int {
 	case de.isParent:
 		namelen = 1
 	default:
-		namelen = len(de.filename)
+		if de.joliet {
+			namelen = len(ucs2StringToBytes(de.filename))
+		} else {
+			namelen = len(de.filename)
+		}
 	}
 
 	return namelen
@@ -118,27 +123,31 @@ func (de *directoryEntry) toBytes(skipExt bool, ceBlocks []uint32) ([][]byte, er
 	case de.isParent:
 		filenameBytes = []byte{0x01}
 	default:
-		// first validate the filename
-		err = validateFilename(de.filename, de.isSubdirectory, de.filesystem.suspEnabled)
-		if err != nil {
-			nametype := "filename"
-			if de.isSubdirectory {
-				nametype = "directory"
+		if de.joliet {
+			filenameBytes = ucs2StringToBytes(de.filename)
+		} else {
+			// first validate the filename
+			err = validateFilename(de.filename, de.isSubdirectory, de.filesystem.suspEnabled)
+			if err != nil {
+				nametype := "filename"
+				if de.isSubdirectory {
+					nametype = "directory"
+				}
+				return nil, fmt.Errorf("invalid %s %s: %v", nametype, de.filename, err)
 			}
-			return nil, fmt.Errorf("invalid %s %s: %v", nametype, de.filename, err)
-		}
-		filenameBytes, err = stringToASCIIBytes(de.filename)
-		if err != nil {
-			return nil, fmt.Errorf("error converting filename to bytes: %v", err)
+			filenameBytes, err = stringToASCIIBytes(de.filename)
+			if err != nil {
+				return nil, fmt.Errorf("error converting filename to bytes: %v", err)
+			}
 		}
 	}
 
 	// copy it over
 	copy(b[33:], filenameBytes)
 
-	// output directory entry extensions - but only if we did not skip it
+	// output directory entry extensions - but only if we did not skip it (Joliet entries never have SUSP)
 	var extBytes [][]byte
-	if !skipExt {
+	if !skipExt && !de.joliet {
 		extBytes, err = dirEntryExtensionsToBytes(de.extensions, directoryEntryMaxSize-len(b), de.filesystem.blocksize, ceBlocks)
 		if err != nil {
 			return nil, fmt.Errorf("enable to convert directory entry SUSP extensions to bytes: %v", err)
@@ -200,6 +209,10 @@ func dirEntryExtensionsToBytes(extensions []directoryEntrySystemUseExtension, ma
 }
 
 func dirEntryFromBytes(b []byte, ext []suspExtension) (*directoryEntry, error) {
+	return dirEntryFromBytesWithJoliet(b, ext, false)
+}
+
+func dirEntryFromBytesWithJoliet(b []byte, ext []suspExtension, joliet bool) (*directoryEntry, error) {
 	// has to be at least 34 bytes
 	if len(b) < int(directoryEntryMinSize) {
 		return nil, fmt.Errorf("cannot read directoryEntry from %d bytes, fewer than minimum of %d bytes", len(b), directoryEntryMinSize)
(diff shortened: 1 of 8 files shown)


luthermonson:
@deitch here we go 🚀

filesystem/iso9660/directoryentry.go
```diff
@@ -53,7 +54,11 @@ func (de *directoryEntry) countNamelenBytes() int {
 	case de.isParent:
 		namelen = 1
 	default:
```
deitch, line 58: This is a strange structure. It already is in a `switch...case` structure. This should just be:

```go
case de.isParent:
    namelen = 1
case de.joliet:
    namelen = len(ucs2StringToBytes(de.filename))
default:
    namelen = len(de.filename)
```

filesystem/iso9660/directoryentry.go
```diff
@@ -118,27 +123,31 @@ func (de *directoryEntry) toBytes(skipExt bool, ceBlocks []uint32) ([][]byte, er
 	case de.isParent:
 		filenameBytes = []byte{0x01}
 	default:
-		// first validate the filename
-		err = validateFilename(de.filename, de.isSubdirectory, de.filesystem.suspEnabled)
-		if err != nil {
-			nametype := "filename"
-			if de.isSubdirectory {
-				nametype = "directory"
+		if de.joliet {
```
deitch: Same as previous comment. This should be just another `case` option.

deitch: Also, does this mean that it will support either Joliet or RR? Those are not mutually exclusive, it should be able to support both.

filesystem/iso9660/directoryentry.go
```diff
@@ -118,27 +123,31 @@ func (de *directoryEntry) toBytes(skipExt bool, ceBlocks []uint32) ([][]byte, er
...
-		if err != nil {
-			return nil, fmt.Errorf("error converting filename to bytes: %v", err)
 		}
 	}
 
 	// copy it over
 	copy(b[33:], filenameBytes)
 
-	// output directory entry extensions - but only if we did not skip it
+	// output directory entry extensions - but only if we did not skip it (Joliet entries never have SUSP)
```
deitch: Is that true? I had thought an ISO could have both RR and Joliet simultaneously. They are stored in different places - RR in SUSP extensions, Joliet in SVD - but they do not conflict.

deitch: Or did I misunderstand how this section works?

filesystem/iso9660/directoryentry.go
```diff
@@ -244,12 +257,16 @@ func dirEntryFromBytes(b []byte, ext []suspExtension) (*directoryEntry, error) {
 		filename = ""
 		isParent = true
 	default:
-		filename = string(nameBytes)
+		if joliet {
```
deitch: Same as earlier. Why have an `if ... then` statement inside a `switch ... case` in the `default` case? Just make it another case.

filesystem/iso9660/finalize.go
```diff
@@ -544,6 +627,10 @@ func (fsm *FileSystem) Finalize(options FinalizeOptions) error {
 	if options.ElTorito != nil {
 		rootLocation++
 	}
+	// if Joliet was enabled, use one sector for the supplementary volume descriptor
+	if options.Joliet {
```
deitch, line 631: How does this connect with the previous one? What if it is ElTorito && Joliet? Is that rootLocation++ twice? Or just once?

filesystem/iso9660/finalize.go
```diff
@@ -667,6 +754,75 @@ func (fsm *FileSystem) Finalize(options FinalizeOptions) error {
 		catEntry.content = bootcat
 	}
 
+	// build Joliet directory tree if enabled
```
deitch, line 763: Why are all of these vars outside of `if options.Joliet`, and then again a few lines down, when the interim lines do not appear to be affected by it? Can these be consolidated into a single `if options.Joliet`, and then the vars are inside the `if` clause?

deitch (changes requested):
Some cleanup suggestions, but mostly some structural and understanding questions.

luthermonson:
sorry about the ai slop, ill get it all fixed. thanks for the feedback

filesystem/iso9660/finalize.go
luthermonson: claude was weirdly specific about this, does this make sense? ElTorito + Joliet rootLocation++ (finalize.go:631): He asks whether both ElTorito and Joliet each bump rootLocation by one sector. The answer is yes — each volume descriptor occupies its own sector (ElTorito boot record VD, then Joliet SVD), so the ++ stacks. But the code doesn't make this obvious, and a brief comment explaining  the sector layout would help.

filesystem/iso9660/directoryentry.go
luthermonson: this is a reply from claude, based on what i know about rr/joliet i believe it...   You're right, they're not mutually exclusive and the code does support both simultaneously. The de.joliet flag is per-directory-entry, not per-filesystem — it marks entries that belong to the Joliet SVD tree. When you finalize with both RockRidge: true and Joliet: true, two separate directory trees get built: the PVD tree gets its normal RR/SUSP extensions, and the Joliet SVD tree gets UCS-2 encoded filenames. Each directoryEntry only ever belongs to one tree, so when the comment says "Joliet entries never have SUSP" it's talking about the SVD directory entries specifically, not the filesystem as a whole. The PVD entries still get full Rock Ridge treatment. I'll clean up the comment and the if/switch structure to make that clearer

luthermonson:
everything refactored, sorry i missed that weird case/if thing claude was doing. the refactor looks much cleaner now, thanks for the feedback

deitch (dismissed):
This looks much cleaner.

Can you run a manual test? Create an iso with RR and Joliet and ElTorito, and then run analysis on it (`xorriso` or whatever tool you like) to see that it has everything? if you did already, just comment here. 

I don't think any of the automated tests cover the "all of the above" case.

luthermonson:
> Can you run a manual test? 

lol no, faster to just have claude do this https://github.com/diskfs/go-diskfs/pull/372/changes/9d16e1ab0b8bb98e50818c699e6d36245c9c3571

luthermonson:
closes #73

deitch (approved):

deitch:
Great!!

rgl:
Does this mean "Joliet extensions to iso9660" can be dropped from the plan section of the readme too?

deitch:
Good point, thanks. I just pulled a PR to update it.

Issue #360, opened 2026-03-31
Squashfs fails to write larger files


Throvn:
When creating a squashfs filesystem using the folder structure below I expect to be able to unsquashfs the filesystem and get the same folder structure back.

```
root/
    a/
        large.bin
    b/
        ...
```

However, everything which would be written after the large.bin binary, won't be written. When trying to unpack the squashfs (using `unsquashfs`), an inode error is thrown.

```
get_metadata: failed to read block
FATAL ERROR: read_inode: failed to read inode 4784488:1512
```
Everything after the binary cannot be unsquashed. However, when trying to execute the binary, it is working as expected, therefore the squashing of the large file itself is successful.

The example is based on the filepath.Walk example, therefore `a/` is walked before `b/`.
FYI, the file is around 10 mb.

Throvn:
Your understanding is correct. Sorry for the long reply period. I'm too busy to recreate it for now. When I come back to the project and still run into the error, I will provide you with a minimum reproducable example.

Until then, I'm not sure what to do with this issue.
If you prefer, you can close it until then.

luthermonson:
i made claude write a test last night and it found some Closes were not being called in the existing tests, it's possible in your code you're doing something similar but tl;dr squashfs works fine if you close and finalize everything

luthermonson:
ah im sorry, i was using sonnet and didn't notice it. opus found a real bug. i'll have a commit up shortly

luthermonson:
sorry for the spam as i'm still getting used to what these models can actually do, there is now a failing tests PR #376 up to prove there is a problem in finalize and the fix is in #377 hope that helps

deitch:
So now that #377 is in, it looks like it includes #376 test? Or do we need to rebase #376 with the fixes of #377 to see if it solves?

luthermonson:
nah, close it

deitch:
Huzzah! Thank you @luthermonson

PR #376, opened 2026-04-16
squashfs: add failing test reproducing issue #360

luthermonson:
Demonstrates the root cause: updateInodeLocations computes inode block byte offsets as `logicalBlock * 8194` assuming every compressed metadata block is exactly 8194 bytes. That's only true with compression off. With real gzip compression blocks are typically much smaller, so directory entries point past the real block positions.

Reproducer uses 600 small files + CompressorGzip{CompressionLevel: 9} to force both conditions (inodes span multiple metadata blocks AND compression actually reduces block size).

This test currently FAILS — all 600 files unreadable, even the root directory lookup fails with "could not read directory entries for .". Existing tests didn't catch this because they use FinalizeOptions{} (no compression) or CompressorGzip{} (level 0 = no compression).

Fix to follow in a separate PR.

### DIFF
--- a/filesystem/squashfs/finalize_issue360_test.go
+++ b/filesystem/squashfs/finalize_issue360_test.go
@@ -50,7 +50,7 @@ func TestFinalizeInodesAcrossMetadataBlocks(t *testing.T) {
 		if err != nil {
 			t.Fatalf("OpenFile %s: %v", name, err)
 		}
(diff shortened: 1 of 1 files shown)


luthermonson:
Failing Tests for #360 can be closed after #377 is merged

PR #377, opened 2026-04-16
squashfs: fix inode block references when metadata is compressed (#360)

luthermonson:
updateInodeLocations pre-computed each inode's metadata-block byte offset as `logicalBlock * (metadataBlockSize + 2)` = `logicalBlock * 8194`, which assumes every compressed metadata block is exactly 8194 bytes on disk. That's only true when compression is effectively off. With real gzip compression metadata blocks are typically much smaller, so directory entries and the superblock root-inode ref pointed past the real block positions. Readers then failed with "could not read directory entries" or "failed to read inode X:Y" for every file whose inode was not in block 0.

The bug reproduces without any "large" file — it just needs enough inodes to span a metadata block boundary AND effective compression. A single large file triggers it too because its extendedFile inode carries one block-size entry per data block (a 1 GB file at 128 KB blocksize needs ~32 KB of inode, spanning 4 metadata blocks by itself).

Fix: store a logical block index in updateInodeLocations, then have writeInodes track the actual on-disk byte offset of each metadata block as it writes them. After writeInodes returns, translateInodeLocations rewrites every inodeLocation.block and every directory entry's startBlock from logical index to real byte offset, before the directory table (which embeds those references) is written.

There's a matching TODO at finalize.go:274 noting "blockPosition calculations appear to be off" — this fixes that.

The reproducer test TestFinalizeInodesAcrossMetadataBlocks was added as a standalone failing test in a preceding PR; it passes after this fix.

### DIFF
--- a/filesystem/squashfs/finalize.go
+++ b/filesystem/squashfs/finalize.go
@@ -1461,7 +1461,7 @@ func updateInodeLocations(files []*finalizeFileInfo) {
 // translateInodeLocations walks every inode location and every directory entry
 // and replaces its logical block index with the actual byte offset of that
 // metadata block in the inode table, using the offsets recorded by writeInodes.
-func translateInodeLocations(files []*finalizeFileInfo, directories []*finalizeFileInfo, blockOffsets []int64) {
+func translateInodeLocations(files, directories []*finalizeFileInfo, blockOffsets []int64) {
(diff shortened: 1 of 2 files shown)


luthermonson:
Fixes #360

deitch (approved):

PR #378, opened 2026-04-21
fix: accept blocksize=4096 in fat32.Create for 4k native block devices

jfroy:
This patch allows creating FAT32 filesystems with 4K block size. This is intended to create UEFI System Partitions (ESPs) on NVME namespaces with a 4Kn format.

This is a follow-up to PR #320 which added support for reading FAT32 filesystems with 4K block size.

### DIFF
--- a/filesystem/fat32/fat32.go
+++ b/filesystem/fat32/fat32.go
@@ -30,6 +30,7 @@ type SectorSize = fat12.SectorSize
 
 const (
 	SectorSize512        SectorSize = 512
+	SectorSize4096       SectorSize = 4096
 	bytesPerSlot         int        = 32
 	maxCharsLongFilename int        = 13
 )
@@ -76,19 +77,25 @@ func (fs *FileSystem) SetLabel(volumeLabel string) error {
 
 // Create creates a FAT32 filesystem on the given backend.
 func Create(b backend.Storage, size, start, blocksize int64, volumeLabel string, reproducible bool) (*FileSystem, error) {
-	if blocksize != int64(SectorSize512) && blocksize > 0 {
-		return nil, fmt.Errorf("blocksize for FAT32 must be either 512 bytes or 0, not %d", blocksize)
+	// Check writability first so a readonly backend surfaces the plain
+	// backend error rather than a layout/size validation error.
+	if _, err := b.Writable(); err != nil {
+		return nil, err
+	}
+	if blocksize != int64(SectorSize512) && blocksize != int64(SectorSize4096) && blocksize > 0 {
+		return nil, fmt.Errorf("blocksize for FAT32 must be either 512 bytes, 4096 bytes, or 0; not %d", blocksize)
+	}
+	if blocksize == 0 {
+		blocksize = int64(SectorSize512)
 	}
 	if size > Fat32MaxSize {
 		return nil, fmt.Errorf("requested size is larger than maximum allowed FAT32, requested %d, maximum %d", size, Fat32MaxSize)
 	}
-	if size < blocksize*4 {
-		return nil, fmt.Errorf("requested size is smaller than minimum allowed FAT32, requested %d minimum %d", size, blocksize*4)
-	}
-	// Check writability early so callers get the plain backend error, not a
-	// wrapped one (preserves the error message that existing tests expect).
-	if _, err := b.Writable(); err != nil {
-		return nil, err
+	// Reserved area: boot sector at 0, FSInfo at 1, their backups at 6 and 7;
+	// rounded up to 32 so the first FAT begins on a 16 KiB boundary at 512 bps.
+	const reservedSectors = uint16(32)
+	if size < int64(reservedSectors)*blocksize {
+		return nil, fmt.Errorf("requested size is smaller than minimum allowed FAT32, requested %d minimum %d", size, int64(reservedSectors)*blocksize)
 	}
 
 	var volid uint32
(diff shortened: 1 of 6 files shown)


jfroy:
I will work on and send tests after an initial review of the main patch.

jfroy:
Looking at the Linux failure. I did the initial work on macOS, not surprised I didn’t catch everything. And then I’ll be submitting new tests for 4k creation.

filesystem/fat32/fat32.go
```diff
@@ -100,25 +104,37 @@ func Create(b backend.Storage, size, start, blocksize int64, volumeLabel string,
...
-		sectorsPerCluster = 32
+		clusterBytes = 16 * KB
 	case size <= 32*GB:
-		sectorsPerCluster = 64
+		clusterBytes = 32 * KB
 	case size <= Fat32MaxSize:
-		sectorsPerCluster = 128
+		clusterBytes = 64 * KB
+	}
+	// FAT32 limits clusters to 32 KiB, but maintain 64 KiB behavior for 512 bps.
```
deitch: I don't understand this comment. If clusters are max 32K, then how does that last `case` make sense?

deitch: Based on the next line, and the comment above, it looks like you are saying, it might be that the filesystem size is > 32GB, sector size != 512, in which case, force it back to 32KB cluster size? So is the logic?

* if sector size is 4K, then cluster bytes can be up to 32K
* if sector size is 512B, then cluster bytes can be up to 64K

Is that it? If not, kindly do explain; if not, do you mind updating the comments to make it clearer?

filesystem/fat32/fat32.go
```diff
@@ -100,25 +104,37 @@ func Create(b backend.Storage, size, start, blocksize int64, volumeLabel string,
...
+	if blocksize != int64(SectorSize512) && clusterBytes > 32*KB {
+		clusterBytes = 32 * KB
+	}
+	sectorsPerCluster := uint8(clusterBytes / blocksize)
+	if sectorsPerCluster == 0 {
```
deitch, line 127: At this point, is it an error if `sectorsPerCluster == 0`? You just set it as:

```go
	sectorsPerCluster := uint8(clusterBytes / blocksize)
```

And earlier, you handled setting `clusterBytes` for every possible case, except if filesystem size > maxFAT32Size, but we handled that on line 86. So how can this line ever resolve as `true`?

deitch (commented):
Overall, this is quite good. Thank you for submitting it. I had only a few questions about things that are unclear in one file.

filesystem/fat32/fat32.go
jfroy: It's an integer division. So if the requested filesystem size is very small, you can get `clusterBytes` < `blocksize`, which will yield zero. I can update this to use `max` instead.

filesystem/fat32/fat32.go
jfroy: 64 KiB clusters are out of spec, but were allowed for 512 bps filesystems before this patch. So the current patch keeps the existing behavior for backward compat but is stricter for 4K bps.

I had Claude do an analysis on mkfs.fat and the behavior of go-diskfs is different. The thresholds are hardcoded as if sector size is 512, but divided into the actual sector count. This means the byte thresholds scale with sector size and mkfs.fat violates the spec for bps > 512:

```
  ┌──────┬──────────────┬───────────────┬─────┬─────────────────────┬──────────────────────┐                                                                   
  │ Tier │ 512B sectors │ 4096B sectors │ spc │ Cluster size (512B) │ Cluster size (4096B) │
  ├──────┼──────────────┼───────────────┼─────┼─────────────────────┼──────────────────────┤                                                                   
  │ 1    │ ≤260 MiB     │ ≤2 GiB        │ 1   │ 512 B               │ 4 KiB                │
  ├──────┼──────────────┼───────────────┼─────┼─────────────────────┼──────────────────────┤
  │ 2    │ ≤8 GiB       │ ≤64 GiB       │ 8   │ 4 KiB               │ 32 KiB               │                                                                   
  ├──────┼──────────────┼───────────────┼─────┼─────────────────────┼──────────────────────┤                                                                   
  │ 3    │ ≤16 GiB      │ ≤128 GiB      │ 16  │ 8 KiB               │ 64 KiB               │                                                                   
  ├──────┼──────────────┼───────────────┼─────┼─────────────────────┼──────────────────────┤                                                                   
  │ 4    │ ≤32 GiB      │ ≤256 GiB      │ 32  │ 16 KiB              │ 128 KiB              │
  ├──────┼──────────────┼───────────────┼─────┼─────────────────────┼──────────────────────┤                                                                   
  │ 5    │ >32 GiB      │ >256 GiB      │ 64  │ 32 KiB              │ 256 KiB              │
  └──────┴──────────────┴───────────────┴─────┴─────────────────────┴──────────────────────┘                                                                   

```

See also https://github.com/dosfstools/dosfstools/blob/289a48b9cb5b3c589391d28aa2515c325c932c7a/src/mkfs.fat.c#L659 and https://download.microsoft.com/download/1/6/1/161ba512-40e2-4cc9-843a-923143f3456c/fatgen103.doc (alt: https://aeb.win.tue.nl/linux/fs/fat/fatgen103.pdf)

Linux's fat driver does not care about bytes per cluster. So the concern is likely for Windows and for UEFI firmware implementations.

filesystem/fat32/fat32.go
deitch: Ah, that is a good point, which I hadn't considered. No, I think this is fine, just add your explanation to the comment right before line 127. That way, some future person doesn't think it is extraneous and remove it, and break something.

filesystem/fat32/fat32.go
deitch: > 64 KiB clusters are out of spec, but were allowed for 512 bps filesystems before this patch. So the current patch keeps the existing behavior for backward compat but is stricter for 4K bps.

Oh, that is just for backwards compatibility for this library? Don't bother with that. Do the right thing. The really nice thing about go modules, is that it is hash-coded by default. So someone using the old non-compliant version only will break if/when they upgrade to something new. Let's not carry spec-breaking things forward.

deitch:
Also small `go test` error.

jfroy:
Resolved comments, added tests, fixed failing test.

deitch (approved):

deitch:
looks great @jfroy, merci. 

I would be happy to hear how you are using it (and I like meeting people in our business). If up for it, let me know; your email is on your profile page, I will drop you a note.

PR #386, opened 2026-04-30
fix(iso9660): handle unspecified timestamps in volume descriptors

m-macdonald:
ECMA-119 Section 9.4.27.2 defines all-zero timestamp fields as "not specified", which is a valid state for all four volume descriptor timestamp fields (creation, modification, expiration, effective).

The expiration and effective fields were already handled, but creation and modification were not, causing a parse error when handling ISOs with unspecified timestamps.

Adds a regression test for PVD parsing with null timestamps.


deitch (approved):
Nice catch, thank you for addressing.

PR #387, opened 2026-05-01
fix(iso9660): include final 4-byte system use entry in parsing

### LINKED ISSUE #127: In certain circumstances, go-diskfs can't read the final SUSP entry
Go-diskfs can miss the final System Use Entry in a System Use Field or a Continuation Area if the final entry is 4 bytes long.

You can see an example of this in Openstep4-Pr1User.iso inside this archive: [https://archive.org/download/NeXTOSIMAGES/NeXT_NEXTSTEP_4.0_PR1_(beta).rar](https://archive.org/download/NeXTOSIMAGES/NeXT_NEXTSTEP_4.0_PR1_%28beta%29.rar).

Try using go-diskfs to list the contents of the rr_moved directory in the above image.

## Expected behavior

The directory appears empty because all directory entries have the RE system use entry.

## Actual behavior
The directory does not appear empty. Go-diskfs misses some RE system use entries, which are 4 bytes in length, when they are the final entry in the system use field.

## The cause

This is likely due to a mistake on [directoryentrysystemuseextension.go:514](https://github.com/diskfs/go-diskfs/blob/e71ece9/filesystem/iso9660/directoryentrysystemuseextension.go#L514).

The line currently reads

```go
for i := 0; i+4 < len(b); {
```

I believe it should read

```go
for i := 0; i+3 < len(b); {
```

A four byte field would take up i+0, i+1, i+2, and i+3, so as long as i+3 is less then len(b), we should be able to read.

I made the same mistake in my own implementation, which is how I found this.

m-macdonald:
The loop condition `i+4 < len(b)` incorrectly excludes a valid 4-byte System Use Field when it falls at the end of a System Use Area, since the 4-byte field occupies bytes i through i+3.

Updated to `i+3 < len(b)`

Fixes #127

### DIFF
--- a/filesystem/iso9660/directoryentrysystemuseextension.go
+++ b/filesystem/iso9660/directoryentrysystemuseextension.go
@@ -511,7 +511,7 @@ func parseDirectoryEntryExtensions(b []byte, handlers []suspExtension) ([]direct
 	entries := make([]directoryEntrySystemUseExtension, 0)
 	lastEntryBySignature := map[string]directoryEntrySystemUseExtension{}
 	// minimum size of 4 bytes for any SUSP entry
-	for i := 0; i+4 < len(b); {
+	for i := 0; i+3 < len(b); {
 		// get the indicator
 		signature := string(b[i : i+2])
 		size := b[i+2]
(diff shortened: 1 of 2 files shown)


deitch (approved):
Particularly appreciate adding the explicit test.

PR #391, opened 2026-05-09
joliet: svd ucs2 volume identifier

luthermonson:
finally had some real world tests and found some identifier issues https://github.com/luthermonson/go-proxmox/issues/258

### DIFF
--- a/filesystem/iso9660/finalize.go
+++ b/filesystem/iso9660/finalize.go
@@ -1172,10 +1172,10 @@ func createJolietPathTable(dirs []*finalizeFileInfo, jolietLocations []uint32) *
 	indexMap := make(map[*finalizeFileInfo]int)
 	entries := make([]*pathTableEntry, 0, len(fis))
 	for i, e := range fis {
-		// Per ECMA-119 §6.9.2 the root directory identifier in the path table
-		// is a single (00) byte even for the Joliet SVD. Other entries keep
-		// their original (UCS-2-encoded-on-disk) name so that lookups like
-		// readDirectoryJoliet("SubDir") can match.
+		// Per ECMA-119 (6th ed.) Annex C.4.6 a) the root directory identifier
+		// in the path table is a single (00) byte even for the Joliet SVD.
+		// Other entries keep their original (UCS-2-encoded-on-disk) name so
+		// that lookups like readDirectoryJoliet("SubDir") can match.
 		var name string
 		if e.isRoot {
 			name = "\x00"
--- a/filesystem/iso9660/finalize.go
+++ b/filesystem/iso9660/finalize.go
@@ -1172,10 +1172,13 @@ func createJolietPathTable(dirs []*finalizeFileInfo, jolietLocations []uint32) *
 	indexMap := make(map[*finalizeFileInfo]int)
 	entries := make([]*pathTableEntry, 0, len(fis))
 	for i, e := range fis {
-		// root uses 0x01, others use original name
+		// Per ECMA-119 §6.9.2 the root directory identifier in the path table
+		// is a single (00) byte even for the Joliet SVD. Other entries keep
+		// their original (UCS-2-encoded-on-disk) name so that lookups like
+		// readDirectoryJoliet("SubDir") can match.
 		var name string
 		if e.isRoot {
-			name = string([]byte{0x01})
+			name = "\x00"
 		} else {
 			name = e.name
 		}
--- a/filesystem/iso9660/pathtable.go
+++ b/filesystem/iso9660/pathtable.go
@@ -138,8 +138,8 @@ func (pt *pathTable) toJolietMBytes() []byte {
 }
 
 // jolietPathTableName returns the UCS-2 encoded name for a path table entry.
-// The root entry (single byte 0x00 per ECMA-119 §6.9.2) stays as-is; all
-// others are UCS-2 encoded.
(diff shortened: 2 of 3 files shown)


filesystem/iso9660/finalize.go
```diff
@@ -1172,10 +1172,13 @@ func createJolietPathTable(dirs []*finalizeFileInfo, jolietLocations []uint32) *
 	indexMap := make(map[*finalizeFileInfo]int)
 	entries := make([]*pathTableEntry, 0, len(fis))
 	for i, e := range fis {
-		// root uses 0x01, others use original name
+		// Per ECMA-119 §6.9.2 the root directory identifier in the path table
```
deitch: This is odd. The [most recent edition](https://ecma-international.org/wp-content/uploads/ECMA-119_6th_edition_december_2025.pdf) does not show it as 6.9.2, but instead in appendix C in C.4.6.a. You still are correct to make it a `00` and not `01`.

filesystem/iso9660/finalize.go
```diff
@@ -1172,10 +1172,13 @@ func createJolietPathTable(dirs []*finalizeFileInfo, jolietLocations []uint32) *
...
+		// is a single (00) byte even for the Joliet SVD. Other entries keep
+		// their original (UCS-2-encoded-on-disk) name so that lookups like
+		// readDirectoryJoliet("SubDir") can match.
 		var name string
 		if e.isRoot {
-			name = string([]byte{0x01})
+			name = "\x00"
```
deitch, line 1181: This is the same as doing `string([]byte{0x00})`, correct? To make sure I am not missing anything, your changes are:

- from 01 -> 00 (per the spec)
- making it a string directly using the escape sequence, rather than the `string([]byte{})` conversion, which does not affect anything

Is that correct?

filesystem/iso9660/pathtable.go
```diff
@@ -230,7 +231,7 @@ func parseJolietPathTable(b []byte) *pathTable {
 		nameBytes := b[i+8 : i+8+int(nameSize)]
 		var name string
 		if nameSize == 1 {
-			name = string(nameBytes) // root entry (0x01)
+			name = string(nameBytes) // root entry: single (00) byte per ECMA-119 §6.9.2
```
deitch: Same comment as before. Where does this actually appear?

filesystem/iso9660/volume_descriptor.go
```diff
@@ -372,8 +372,8 @@ func parseSupplementaryVolumeDescriptor(b []byte) (*supplementaryVolumeDescripto
 
 	return &supplementaryVolumeDescriptor{
 		volumeFlags:                b[7],
-		systemIdentifier:           string(b[8:40]),
-		volumeIdentifier:           string(b[40:72]),
+		systemIdentifier:           bytesToUCS2String(b[8:40]),
+		volumeIdentifier:           bytesToUCS2String(b[40:72]),
```
deitch: Is this correct? UCS2 is used for Joliet, but Joliet is not the only usage of an SVD. Are there cases where this might be incorrect?

deitch: Yes, I do realize that you are doing it symmetrically, i.e. `bytesToUCS2String()` on the `parse()`, and `ucs2StringToBytes()` on the `toBytes()`, but I do not know if that assumption would break something else (now or future).

deitch (commented):
Some questions

filesystem/iso9660/finalize.go
luthermonson: correct

filesystem/iso9660/volume_descriptor.go
luthermonson: ya im sorry i dont actually know, this did make joliet + windows actually work though. i can refactor if you think this is only a joliet problem

filesystem/iso9660/finalize.go
luthermonson: comment from claude, i dunno where it found it but i can make it fix the citation

filesystem/iso9660/pathtable.go
luthermonson: ill track it down

filesystem/iso9660/volume_descriptor.go
deitch: I really am not sure. I think to be safe, we should make it Joliet-only, unless that seriously complicates it.

filesystem/iso9660/finalize.go
deitch: OK, no problem, then.

filesystem/iso9660/finalize.go
deitch: Thanks

filesystem/iso9660/pathtable.go
deitch: Thanks

filesystem/iso9660/volume_descriptor.go
luthermonson: new commit has this change

luthermonson:
i think i have addressed all your comments

deitch (approved):
Thanks for the patience.

luthermonson:
no worries at all, can i get a v1.9.3 so i can pin this and get go-proxmox buttoned up?

deitch:
Sure.

PR #392, opened 2026-05-12
fix(iso9660): handle zero-filled system use area padding in directory entries

### LINKED ISSUE #388: Panic in parseSystemUseExtensionRaw when parsing DVD ISOs with zero-filled system use area padding
It seems commercially produced DVD ISOs often have zero-filled padding in the system use area which ultimately causes a panic in `parseSystemUseExtensionRaw` as it receives an empty slice.

I believe zero-filled padding is technically non-conformant per the SUSP spec, as it requires a minimum 4-byte header for all system use fields. However, this pattern is consistent across the commercial DVD titles I've tested and I think it would be preferable to handle it gracefully rather than panicking.

I think `parseDirectoryEntryExtensions` would benefit from a check within the loop to make sure that the size is at least 4 bytes
[directoryentrysystemuseextension.go:L517](https://github.com/diskfs/go-diskfs/blob/e59f4c75f36e66d6ef4c27b097c16570eb7fc0a3/filesystem/iso9660/directoryentrysystemuseextension.go#L517)
```go
size := b[i+2]
// Add a check for size
if size < 4 {
    break
}
suspBytes := b[i : i+int(size)]
```

I would be more than happy to open a PR with a fix if this is something worth addressing.

m-macdonald:
Added a check in parseDirectoryEntryExtensions to treat any entry with a length less than 4 as the end of data, aligning with the SUSP spec's minimum 4-byte header requirement.

Fixes #388

### DIFF
--- a/filesystem/iso9660/directoryentrysystemuseextension.go
+++ b/filesystem/iso9660/directoryentrysystemuseextension.go
@@ -515,6 +515,9 @@ func parseDirectoryEntryExtensions(b []byte, handlers []suspExtension) ([]direct
 		// get the indicator
 		signature := string(b[i : i+2])
 		size := b[i+2]
+		if size < 4 {
+			break
+		}
 		suspBytes := b[i : i+int(size)]
 		var (
 			entry directoryEntrySystemUseExtension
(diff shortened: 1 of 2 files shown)


deitch (approved):
Thanks for this.

PR #393, opened 2026-05-14
ext4: expose inode metadata via FileInfo.Sys()

### LINKED ISSUE #301: [Feature request] expose filesystem inode information
Hi!

Thanks for this package, really useful!

Currently, [filesystem.Stat](https://pkg.go.dev/github.com/diskfs/go-diskfs@v1.7.0/filesystem/ext4#FileSystem.Stat) returns the information of a path in a fs.FileInfo struct: https://pkg.go.dev/github.com/diskfs/go-diskfs@v1.7.0/filesystem/ext4#FileSystem.Stat 

This struct doesn't offer all the fields that for example the ext4 format has, like the inode.

There're some structs that could expose their internal fields:

https://github.com/diskfs/go-diskfs/blob/63be039c8b725fad4d0666ae0d381ff3907b47bd/filesystem/ext4/directory.go#L15
https://github.com/diskfs/go-diskfs/blob/63be039c8b725fad4d0666ae0d381ff3907b47bd/filesystem/ext4/directoryentry.go#L27

https://github.com/diskfs/go-diskfs/blob/63be039c8b725fad4d0666ae0d381ff3907b47bd/filesystem/ext4/file.go#L10-L11
https://github.com/diskfs/go-diskfs/blob/master/filesystem/ext4/inode.go#L110

For example, the `stat` command shows much more information than what the FileInfo struct offers:

```bash
~ $ stat .
  File: .
  Size: 4096      	Blocks: 8          IO Block: 4096   directory
 Device: 253,3	Inode: 3977042     Links: 7
 Access: (0775/drwxrwxr-x)  Uid: ( 1000/      ga)   Gid: ( 1000/      ga)
 Context: unconfined_u:object_r:user_home_t:s0
 Access: 2025-09-30 23:17:38.833369333 +0200
 Modify: 2025-09-30 23:17:37.589361321 +0200
 Change: 2025-09-30 23:17:37.589361321 +0200
 Birth: 2025-09-26 00:21:16.319751387 +0200
```

More like Stat_t: https://pkg.go.dev/golang.org/x/sys/unix#Stat_t

For some applications, having access to this information would be quite useful.

luthermonson:
Closes #301.

## What this does

`filesystem/ext4.FileSystem.Stat()` returns an `fs.FileInfo` whose `Sys()` previously exposed only `UID`/`GID` (and, as of recent upstream changes, `Major`/`Minor`/`Ino`/`Nlink`). The on-disk ext4 inode carries much more — block count, access/change/create timestamps, symlink target, and a flag bitmap — and this PR surfaces all of it through the existing `Sys()` escape hatch, matching the idiomatic Go pattern used by `os.Stat` returning `*syscall.Stat_t`.

Changes are **additive** — the existing fields on `ext4.StatT` are unchanged, and `FileSystem.Stat`'s signature is unchanged. No change to the `filesystem.FileSystem` interface.

### `StatT` now carries

| Field | Source | Added in |
|---|---|---|
| `UID`, `GID` | `inode.owner` / `inode.group` | (already there) |
| `Major`, `Minor` | `inode.deviceNumber()` | (already there) |
| `Ino` | `inode.number` | (already there) |
| `Nlink` | `inode.hardLinks` | (already there) |
| `Blocks uint64` | `inode.blocks` | this PR |
| `AccessTime time.Time` | `inode.accessTime` | this PR |
| `ChangeTime time.Time` | `inode.changeTime` | this PR |
| `CreateTime time.Time` | `inode.createTime` | this PR |
| `LinkTarget string` | `inode.linkTarget` (symlinks) | this PR |
| `Flags InodeFlags` | curated subset of `inodeFlags` | this PR |

A new exported `InodeFlags` struct exposes the practically useful flag bits — Immutable, AppendOnly, NoDump, NoAtime, Synchronous, Compressed, Encrypted, HashedIndexes, JournalData, HugeFile, Extents, ExtendedAttributes, InlineData, TopDirectory. The internal `inodeFlags` struct stays unexported.

The existing `(i *inode).stat()` helper is extended to populate the new fields, so all three `FileInfo` construction sites (`FileSystem.Stat`, `File.Stat`, `directoryEntryInfo.Info`) continue to use a single source of truth.

### Usage

```go
fi, _ := fs.Stat("/some/path")
if st, ok := fi.Sys().(*ext4.StatT); ok {
    fmt.Println(st.Ino, st.Nlink, st.CreateTime, st.Flags.Immutable)
    if st.LinkTarget != "" {
        fmt.Println("symlink ->", st.LinkTarget)
    }
}
```

### Tests

New `TestStatSysInodeMetadata` exercises `/random.dat`, `/foo`, `/`, and `/symlink.dat` from the existing test image — asserts non-zero inode, plausible timestamps, stable values across repeated `Stat` calls, and non-empty `LinkTarget` for symlinks.

---

## Follow-up PRs

If this lands, three sibling PRs apply the same `Sys()`-based pattern to the other read-supporting filesystems. Each is independent and reviewable on its own.

### iso9660 — [#394](https://github.com/diskfs/go-diskfs/pull/394) (draft)

`directoryEntry.Sys()` currently returns `*RockRidgeInfo` when Rock Ridge is present, otherwise `nil`. The follow-up replaces `RockRidgeInfo` with a richer `iso9660.StatT` always returned by `Sys()`, carrying ISO9660 fields (`Location`, `VolumeSequence`, `IsHidden`, `IsAssociated`, `ExtAttrSize`, `HasExtendedAttrs`, `HasOwnerGroupPermissions`) plus Rock Ridge fields parsed from PX and SL records (`UID`, `GID`, `NLink`, `Inode`, `LinkTarget`, `RockRidge bool`). RR fields stay zero-valued when no extensions are present. This is a breaking change for any caller currently asserting `*RockRidgeInfo`; the in-repo test assertions are updated as part of the PR.

### squashfs — [#395](https://github.com/diskfs/go-diskfs/pull/395) (draft)

squashfs is the outlier: `Sys()` already returns `*directoryEntry` (aliased as the public `squashfs.FileStat`), which already exposes `UID()`, `GID()`, `Xattrs()`, `Readlink()` as accessor methods. To stay consistent with that existing public surface, the follow-up adds two more accessors — `Inode() uint32` and `InodeType() string` — rather than introducing a parallel `StatT` struct. No breaking change to the existing `FileStat` type.

#### The architectural picture

The recent FAT12/16 support PR (#358) restructured the FAT codebase: the actual implementation now lives in `filesystem/fat12`, and `filesystem/fat32` + `filesystem/fat16` are thin wrappers that embed `*fat12.FileSystem` and override `Type()`. Public API parity with the old fat32 surface is preserved through Go type aliases — for example, `type SectorSize = fat12.SectorSize` already exists in `fat32.go`. A type alias (`A = B`) is *identity*, not a new named type, so `fat32.SectorSize` and `fat12.SectorSize` are literally the same type.

The follow-up applies that same pattern for the new `StatT`. The implementation defines `fat12.StatT` once and a `(*directoryEntry).stat()` helper that populates it from `parseDirEntries` data already on the entry. Then:

```go
// filesystem/fat12/fileinfo.go — single source of truth
type StatT struct {
    ReadOnly, Hidden, System, Archive, VolumeLabel bool
    CreateTime, AccessTime                          time.Time
    Cluster                                         uint32
}

// filesystem/fat32/fat32.go
type StatT = fat12.StatT

// filesystem/fat16/fat16.go
type StatT = fat12.StatT
```

#### Why this is the right shape (not three separate copies)

1. **One implementation, three packages covered.** `directoryEntry.stat()` is defined once. `fat16` and `fat32` inherit working `Sys()` automatically because they reuse `fat12`'s `directoryEntry.Info()` and `File.Stat()` code paths.

2. **Same type across packages, by design.** Because `type StatT = fat12.StatT` is an alias (not a new named type with the same field shape), a value of `*fat32.StatT` *is* a `*fat12.StatT`. A downstream library that wants to handle any FAT variant can type-assert against `*fat12.StatT` once and have it match all three — same way `*syscall.Stat_t` matches any UNIX filesystem reachable through `os.Stat`. Callers using a specific FAT package still write the natural import (`fi.Sys().(*fat32.StatT)` works for fat32 users) without forcing them to learn about the fat12 base package.

3. **No drift risk.** If FAT16 later needs a new attribute exposed, it goes in `fat12.StatT` and the other two get it for free. The alternative (three separate `StatT` structs with the same field shape) would inevitably diverge.

4. **Mirrors the upstream architectural choice already made.** This isn't introducing a new pattern; `fat16`/`fat32` are already wrappers around `fat12.FileSystem` and their public types are already aliases. Adding `StatT` as an alias is consistent with how `SectorSize` works today — anything else would create two competing conventions in the same package.

#### Fields

All populated from data `parseDirEntries` already extracts; no new parsing. Attribute bits (`ReadOnly`, `Hidden`, `System`, `Archive`, `VolumeLabel`), the two FAT timestamps that don't map to `ModTime()` (`CreateTime`, `AccessTime`), and the file's starting `Cluster`.

### DIFF
--- a/filesystem/ext4/ext4_test.go
+++ b/filesystem/ext4/ext4_test.go
@@ -808,3 +808,104 @@ func TestChown(t *testing.T) {
 		}
 	})
 }
+
+func TestStatSysInodeMetadata(t *testing.T) {
+	f, err := os.Open(imgFile)
+	if err != nil {
+		t.Fatalf("Error opening test image: %v", err)
+	}
+	defer f.Close()
+
+	b := file.New(f, true)
+	fs, err := Read(b, 100*MB, 0, 512)
+	if err != nil {
+		t.Fatalf("Error reading filesystem: %v", err)
+	}
+
+	tests := []struct {
+		name string
+		path string
+	}{
+		{"regular file", "/random.dat"},
+		{"directory", "/foo"},
+		{"root", "/"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			fi, err := fs.Stat(tt.path)
+			if err != nil {
+				t.Fatalf("Stat failed: %v", err)
+			}
+			sys := fi.Sys()
+			if sys == nil {
+				t.Fatalf("Sys() returned nil")
+			}
+			stat, ok := sys.(*StatT)
+			if !ok {
+				t.Fatalf("Sys() did not return *StatT, got %T", sys)
+			}
+			if stat.Ino == 0 {
+				t.Errorf("expected non-zero inode number, got 0")
+			}
+			if stat.Nlink == 0 {
+				t.Errorf("expected non-zero hardlink count, got 0")
+			}
+			// modify time on FileInfo should match the inode-derived value through Stat
+			if fi.ModTime().IsZero() {
+				t.Errorf("ModTime is zero")
+			}
+			// timestamps must be plausible: not zero, and not absurdly in the future
+			if stat.AccessTime.IsZero() {
+				t.Errorf("AccessTime is zero")
+			}
+			if stat.ChangeTime.IsZero() {
+				t.Errorf("ChangeTime is zero")
+			}
+			upper := time.Now().AddDate(100, 0, 0)
+			lower := time.Date(1990, 1, 1, 0, 0, 0, 0, time.UTC)
+			if stat.AccessTime.After(upper) || stat.AccessTime.Before(lower) {
+				t.Errorf("AccessTime %v outside plausible range", stat.AccessTime)
+			}
+			if stat.ChangeTime.After(upper) || stat.ChangeTime.Before(lower) {
+				t.Errorf("ChangeTime %v outside plausible range", stat.ChangeTime)
+			}
+			// Confirm Stat returns stable values across repeated calls.
+			fi2, err := fs.Stat(tt.path)
+			if err != nil {
+				t.Fatalf("second Stat failed: %v", err)
+			}
+			stat2, ok := fi2.Sys().(*StatT)
+			if !ok {
+				t.Fatalf("second Sys() did not return *StatT")
+			}
+			if stat.UID != stat2.UID || stat.GID != stat2.GID {
+				t.Errorf("UID/GID not stable: %d:%d vs %d:%d", stat.UID, stat.GID, stat2.UID, stat2.GID)
+			}
+			if stat.Ino != stat2.Ino {
+				t.Errorf("inode number not stable: %d vs %d", stat.Ino, stat2.Ino)
+			}
+		})
+	}
+
+	t.Run("symlink target", func(t *testing.T) {
+		// On the test image, /symlink.dat is a symlink to random.dat. fs.Stat returns the
+		// link inode's metadata, which carries LinkTarget for symlinks.
+		// We look up the symlink via the directory entry path since fs.Stat may follow it.
+		_, entry, err := fs.getEntryAndParent("/symlink.dat")
+		if err != nil {
+			t.Fatalf("getEntryAndParent failed: %v", err)
+		}
+		if entry == nil {
+			t.Fatalf("symlink entry not found")
+		}
+		in, err := fs.readInode(entry.inode)
+		if err != nil {
+			t.Fatalf("readInode failed: %v", err)
+		}
+		s := in.stat()
+		if in.fileType == fileTypeSymbolicLink && s.LinkTarget == "" {
+			t.Errorf("expected non-empty LinkTarget for symlink inode")
+		}
+	})
+}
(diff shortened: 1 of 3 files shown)


deitch (approved):

deitch:
This is good. I don't like the PR comment much. I know it is AI generated, and just don't care. It is good and detailed. But it references the follow-on PRs too much, and describes the logic of those in this comment. Not going to hold it up for that, though.

What agent are you using and with which skills, such that it not only creates the comments, but also includes links to PRs?

luthermonson:
> This is good. I don't like the PR comment much. I know it is AI generated, and just don't care. It is good and detailed. But it references the follow-on PRs too much, and describes the logic of those in this comment. Not going to hold it up for that, though.
> 
> What agent are you using and with which skills, such that it not only creates the comments, but also includes links to PRs?

claude opus 4.7

PR #394, opened 2026-05-14
iso9660: expose rich metadata via FileInfo.Sys()

luthermonson:
Refs #301. Sibling of #393 (ext4) — same pattern, applied to iso9660.

## What this does

`filesystem/iso9660/directoryEntry.Sys()` previously returned `*RockRidgeInfo` when Rock Ridge extensions were present on the entry, otherwise `nil`. That was a partial answer to #301: callers could only see uid/gid/mode/symlink, only on Rock Ridge ISOs, and the plain ISO9660 metadata (extent location, volume sequence, hidden flag, etc.) was invisible.

This PR replaces `RockRidgeInfo` with a richer `iso9660.StatT` that `Sys()` always returns, carrying both the plain ISO9660 directory-entry fields and the Rock Ridge fields when present.

### Fields

| Source | Field |
|---|---|
| ISO9660 directory entry | `ExtAttrSize uint8` |
| | `Location uint32` (extent LBA) |
| | `VolumeSequence uint16` |
| | `IsHidden bool` |
| | `IsAssociated bool` |
| | `HasExtendedAttrs bool` |
| | `HasOwnerGroupPermissions bool` |
| Rock Ridge PX record | `UID uint32` |
| | `GID uint32` |
| | `NLink uint32` |
| | `Inode uint32` |
| Rock Ridge SL record | `LinkTarget string` |
| (any RR record found) | `RockRidge bool` |

RR fields are zero-valued / `RockRidge=false` on non-RR ISOs.

### Breaking change — `RockRidgeInfo` removed

`RockRidgeInfo` was added recently (in #353) and exported, so removing it is a breaking change. In-repo callers in 4 test files are updated as part of this PR. External callers will need to switch:

```go
// before
if rri, ok := info.Sys().(*iso9660.RockRidgeInfo); ok {
    fmt.Println(rri.Symlink, rri.UID)
}

// after
if st, ok := info.Sys().(*iso9660.StatT); ok && st.RockRidge {
    fmt.Println(st.LinkTarget, st.UID)
}
```

The alternative — keeping `RockRidgeInfo` alongside the new `StatT` — would have left two competing return types for `Sys()` and split the metadata model in two. Given that `RockRidgeInfo` is fresh (one release) and only covers a subset of what callers actually want, supersession seems cleaner than coexistence. Open to keeping it as a deprecated alias if reviewers prefer.

### Tests

New `TestSysStatT` (plain ISO via `testdata/9660.iso`) and `TestSysStatTRockRidge` (RR ISO via `testdata/rockridge.iso`). Existing Rock Ridge tests (`combined_xorriso_test.go`, `finalize_test.go`, `rockridge_xorriso_test.go`) updated to assert on `*StatT` instead of `*RockRidgeInfo`.

### DIFF
--- a/filesystem/iso9660/combined_xorriso_test.go
+++ b/filesystem/iso9660/combined_xorriso_test.go
@@ -141,10 +141,10 @@ func TestCombinedRockRidgeJolietElTorito(t *testing.T) {
 			if info.Mode()&os.ModeSymlink == 0 {
 				t.Errorf("link.txt: expected symlink mode, got %v", info.Mode())
 			}
-			if rri, ok := info.Sys().(*iso9660.RockRidgeInfo); !ok {
-				t.Error("link.txt: Sys() did not return *RockRidgeInfo")
-			} else if rri.Symlink != "target.txt" {
-				t.Errorf("link.txt: symlink target %q, want %q", rri.Symlink, "target.txt")
+			if stat, ok := info.Sys().(*iso9660.StatT); !ok {
+				t.Error("link.txt: Sys() did not return *StatT")
+			} else if stat.LinkTarget != "target.txt" {
+				t.Errorf("link.txt: symlink target %q, want %q", stat.LinkTarget, "target.txt")
 			}
 		}
 	}
--- a/filesystem/iso9660/directoryentry.go
+++ b/filesystem/iso9660/directoryentry.go
@@ -560,6 +560,36 @@ func (de *directoryEntry) Sys() interface{} {
 	return de.statT()
 }
 
+func (de *directoryEntry) statT() *StatT {
+	s := &StatT{
+		ExtAttrSize:              de.extAttrSize,
+		Location:                 de.location,
+		VolumeSequence:           de.volumeSequence,
+		IsHidden:                 de.isHidden,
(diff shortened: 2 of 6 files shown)


deitch:
This is fine and makes sense. I have no issue with deprecating `RockRidgeInfo`. Once again, go module structure makes life easier. 

2 issues:
1. This is still marked as draft; is it ready?
2. My only request - and it is minor - is that the member func `func (de *directoryEntry) statT() *StatT` is in `fileinfo.go` rather than in `directoryentry.go`. I know, there is no "law" that requires all members to be in a single file, and we probably have places where there are so many member funcs that we had to split it, but I don't think that is the case here, so would rather have it in `directoryentry.go` with the struct and the other funcs. I don't care if we leave `fileinfo.go` with just the `type StatT struct` in it and nothing else, or move that into `directoryentry.go`.

luthermonson:
I will address your feedback later tonight

luthermonson:
Moved `statT()` into `directoryentry.go` next to `Sys()` in a212379. Left the `StatT` struct in `fileinfo.go` (the second of the two options you offered). Ready for another look.

deitch (approved):

PR #395, opened 2026-05-14
squashfs: expose inode number and type via Sys()

luthermonson:
Refs #301. Sibling of #393 (ext4) — same goal, but adapted to squashfs's existing public surface.

## What this does

The other filesystems answering #301 (ext4, iso9660, fat) introduce a per-package `StatT` struct returned by `Sys()`. Squashfs is different: `Sys()` already returns `*directoryEntry` (aliased as the public `squashfs.FileStat`), which already exposes filesystem-specific metadata via accessor methods — `UID()`, `GID()`, `Xattrs()`, `Readlink()`.

To stay consistent with that existing public surface — and to avoid breaking any caller already doing `sys.(squashfs.FileStat).UID()` — this PR adds two more accessors rather than introducing a parallel struct:

- `(*directoryEntry).Inode() uint32` — squashfs inode number
- `(*directoryEntry).InodeType() string` — human-readable inode type (`"basic-file"`, `"extended-symlink"`, etc., one of 14 values)

`Sys()`'s return type is unchanged. Callers use it the same way they already use `UID()`/`GID()`:

```go
if st, ok := fi.Sys().(squashfs.FileStat); ok {
    fmt.Println(st.Inode(), st.InodeType(), st.UID())
}
```

### Why not a `StatT` struct here

Two competing public shapes inside one package (`squashfs.FileStat` accessor methods + `squashfs.StatT` struct fields, both reachable from the same `Sys()` call) would be worse than mild inconsistency across packages. The existing pattern was set when squashfs first exposed UID/GID/Xattrs; this PR follows it.

### Tests

New `TestSquashfsInodeMetadata` exercises `ReadDir(".")` + `DirEntry.Info()` and `OpenFile().Stat()` on the existing test image, asserts non-zero `Inode()`, no duplicate inode numbers across entries, and a sensible `InodeType()` (with the extra check that directories report `basic-directory` or `extended-directory`).

### DIFF
--- a/filesystem/squashfs/directoryentry.go
+++ b/filesystem/squashfs/directoryentry.go
@@ -145,6 +145,53 @@ func (d *directoryEntry) Xattrs() map[string]string {
 	return d.xattrs
 }
 
+// Inode returns the squashfs inode number for this entry.
+func (d *directoryEntry) Inode() uint32 {
+	if d.inode == nil {
+		return 0
+	}
+	return d.inode.index()
+}
+
+// InodeType returns a human-readable name for the squashfs inode type.
+func (d *directoryEntry) InodeType() string {
+	if d.inode == nil {
+		return "unknown"
+	}
+	switch d.inode.inodeType() {
+	case inodeBasicDirectory:
+		return "basic-directory"
+	case inodeBasicFile:
+		return "basic-file"
+	case inodeBasicSymlink:
+		return "basic-symlink"
+	case inodeBasicBlock:
+		return "basic-block-device"
+	case inodeBasicChar:
+		return "basic-char-device"
+	case inodeBasicFifo:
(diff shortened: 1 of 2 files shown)


deitch (approved):
This looks good to me. I am not the biggest fan of the issue you raised - StatT everywhere else, just relying on the `directoryEntry` itself with exposed properties here - but it is good enough. The primary interfaces will work, and we can clean it up later.

If you think it is an issue, we can restructure this to a `StatT` as well. I am not sure how complicated that would be.

deitch:
Also, this still is draft

luthermonson:
i think i'd like to see this structured as a `StatT`, PRs are free so let me open a new one with that design and we compare what we want to go with.

PR #404, opened 2026-05-28
squashfs: honor partition offset in Read and Finalize

eriknordmark:
## squashfs: honor the partition offset in Read and Finalize

`squashfs.Create` and `squashfs.Read` use offsets relative to the start
of the filesystem throughout (superblock at offset 0, then the fragment
/ inode / xattr tables at squashfs-internal offsets parsed from the
superblock). But the `backend.Storage` they were handed always referred
to the whole disk. When the caller passes a non-zero `start` — i.e. the
filesystem lives inside a partition:

- **`Finalize`** wrote every section, including the superblock at
  `WriteAt(sbBytes, 0)`, to absolute backend offsets, clobbering the
  protective MBR / GPT and placing the squashfs in the wrong spot.
- **`Read`** read the superblock at `start` but then dispatched
  `readFragmentTable` / `readXattrsTable` / `readUidsGids` at unbiased
  squashfs-internal offsets, so on a non-zero start it read garbage
  (`error reading fragments: ... unexpected EOF`).

This wraps the backend in `backend.Sub(b, start, size)` at `Create`- and
`Read`-time when `start != 0`. `Sub` auto-biases every `ReadAt`/`WriteAt`
by the partition offset, so the existing internal offset arithmetic is
unchanged — no per-callsite edits. Whole-disk callers (`start == 0`) are
unaffected.

### Test

Adds `TestSquashfsInPartition`: it builds a squashfs inside a GPT
partition 1 MiB in, finalizes, asserts the GPT survived and the
superblock landed at the partition offset, then reopens and reads the
marker file back. It fails without this fix (superblock lands at offset
0); the existing whole-disk tests never caught this because they always
pass `start == 0`.

### Notes

- The `sync/verifyBlockCopy` change that was originally bundled here has
  been split into #406 per review feedback.
- Related to #401 (GPT crash-safety) — independent, can land in either
  order.

### DIFF
--- a/filesystem/squashfs/squashfs.go
+++ b/filesystem/squashfs/squashfs.go
@@ -27,7 +27,6 @@ type FileSystem struct {
 	workspace  string
 	superblock *superblock
 	size       int64
-	start      int64
 	backend    backend.Storage
 	blocksize  int64
 	compressor Compressor
@@ -104,7 +103,6 @@ func Create(b backend.Storage, size, start, blocksize int64) (*FileSystem, error
 	// there is nothing in there
 	return &FileSystem{
 		workspace: tmpdir,
-		start:     0,
 		size:      size,
 		backend:   b,
 		blocksize: blocksize,
@@ -231,7 +229,6 @@ func Read(b backend.Storage, size, start, blocksize int64) (*FileSystem, error) 
 
 	fs := &FileSystem{
 		workspace:  "", // no workspace when we do nothing with it
-		start:      0,  // backend is already biased by start (if non-zero) via Sub above
 		size:       size,
 		backend:    b,
 		superblock: s,
--- a/filesystem/squashfs/squashfs.go
+++ b/filesystem/squashfs/squashfs.go
@@ -90,11 +90,21 @@ func Create(b backend.Storage, size, start, blocksize int64) (*FileSystem, error
 		return nil, fmt.Errorf("could not create working directory: %v", err)
 	}
 
+	// Wrap the backend so all internal ReadAt/WriteAt calls in this
+	// package use offsets relative to the start of the filesystem.
+	// Finalize and Read are coded throughout in terms of squashfs-
+	// internal offsets; without this wrapping, every write/read on a
+	// non-zero start (i.e. inside a partition) would land at the wrong
+	// place on the underlying disk.
+	if start != 0 {
+		b = backend.Sub(b, start, size)
+	}
+
 	// create root directory
 	// there is nothing in there
 	return &FileSystem{
 		workspace: tmpdir,
-		start:     start,
+		start:     0,
 		size:      size,
 		backend:   b,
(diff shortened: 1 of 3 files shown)


filesystem/squashfs/squashfs.go
```diff
@@ -211,7 +231,7 @@ func Read(b backend.Storage, size, start, blocksize int64) (*FileSystem, error)
 
 	fs := &FileSystem{
 		workspace:  "", // no workspace when we do nothing with it
-		start:      start,
+		start:      0,  // backend is already biased by start (if non-zero) via Sub above
```
deitch: I don't quite understand. The current approach is to take the backend "as is", save the `start` in the `FileSystem{}`, and then offset all calls with `start`. Why is it better to offset the backend and set the `start` to 0? Isn't it the same thing? And with this change, don't you need to remove all references to `fs.start` (and maybe remove it as the property entirely)? Or are you saying that the current approach should do that, but ignores `fs.start` anyways?

sync/verify.go
```diff
@@ -21,26 +21,37 @@ func verifyBlockCopy(d *disk.Disk, from, to int, expectedSize int64) error {
...
+	// Hash the leading expectedSize bytes of both partitions and
+	// compare. NewLimitWriter caps the hasher's input, and
+	// origPart.ReadContents would otherwise feed all of the source
+	// partition's bytes into the hasher — for our purposes, we want
+	// only expectedSize bytes hashed on each side regardless of how
+	// large either partition is.
 	origHasher := sha256.New()
-	size, err := origPart.ReadContents(d.Backend, origHasher)
-	if err != nil {
+	if _, err := origPart.ReadContents(d.Backend, NewLimitWriter(origHasher, expectedSize)); err != nil {
```
deitch: These verify changes look fine, but really should be its own PR

deitch (changes requested):
Two comments.

sync/verify.go
eriknordmark: Moving them out to a separate PR.

filesystem/squashfs/squashfs.go
eriknordmark: Apparently there are about 15 places where start is not added in calls to  ReadAt/WriteAt. But more challenging, there are two places where it must not be added (finalize.go:380 and finalize.go:678) since those are ReatAt's from the source.
Thus instead of doing this correctly at each call site it is a lot more clear to bias in the object so that the destination (fs.backend.Writable()) gets the start bias.

FWIW this was originally discovered using a test in partitionresizer, but adding a specific test in this PR to make it be self-contained.

eriknordmark: It turns out that the FileSystem.start was never read anywhere. 
As an optional additional commit in this PR it can be removed.

deitch (approved):

PR #407, opened 2026-05-29
ext4: scale default block size by filesystem size

### LINKED ISSUE #402: ext4: CreateFilesystem fails on multi-GB partitions ("cannot allocate more than 65535 blocks in a single extent")
## Summary

`Disk.CreateFilesystem` with `FSType: filesystem.TypeExt4` fails on partitions larger than ~256 MiB with:

> could not initialize journal: cannot allocate more than 65535 blocks in a single extent

The error originates at `filesystem/ext4/ext4.go:2283`:

```go
if extraBlockCount > maxUint16 {
    return nil, fmt.Errorf("cannot allocate more than %d blocks in a single extent", maxUint16)
}
```

For a 4 KiB-block filesystem this caps a single allocation at 65535 × 4 KiB ≈ 256 MiB. Journal sizing for larger filesystems exceeds that, so creation aborts.

## Reproducer

The cleanest repro I've seen is `diskfs/partitionresizer`'s `TestRun` (https://github.com/diskfs/partitionresizer), which grows existing 500 MiB ext4 partitions to 2 GiB and calls `Disk.CreateFilesystem` for each grown partition. It fails on the first 2 GiB ext4 create:

```
... copying data from original partition 2 to new partition 5
Run failed: failed to create ext4 filesystem for new partition parta: could not initialize journal: cannot allocate more than 65535 blocks in a single extent
```

A minimal standalone repro would be:

```go
d, _ := diskfs.Open("/tmp/2gb.img", ...)
_, err := d.CreateFilesystem(disk.FilesystemSpec{
    Partition: 1,
    FSType:    filesystem.TypeExt4,
})
// err: "could not initialize journal: cannot allocate more than 65535 blocks in a single extent"
```

against a disk image where partition 1 is 2 GiB.

## Observations

The surrounding loop at `ext4.go:2270-2305` already iterates over block groups and appends to an `extents` slice, suggesting multi-extent allocation was intended. The early `extraBlockCount > maxUint16` check at line 2282 short-circuits that before the multi-extent path can be exercised. Capping `extentLength` at `maxBlocksPerExtent` per iteration (which the code already does at line 2295) and letting the outer loop accumulate multiple extents would seem to be the intended shape.

I don't have a tested fix to attach, but happy to follow up if a maintainer confirms the intended approach.

## Impact

Any caller that needs to create an ext4 filesystem larger than ~256 MiB via go-diskfs is blocked. Tools that grow partitions and re-format (such as partitionresizer) inherit the failure.

eriknordmark:
## Summary

`recalculateBlocksize` was a placeholder: every bracket of its size-based
switch set the same `sectorsPerBlock = 2`, so a default-parameter
`Create` always used 1 KiB blocks regardless of filesystem size. On
filesystems of a few GiB the journal allocator — capped at 128 MiB =
131072 1-KiB blocks — asks for more blocks than the per-extent cap
(`maxBlocksPerExtent = 32768`) or the inode's 4-extent root limit can
fit, and `Create` fails with

```
could not initialize journal: cannot allocate more than 65535 blocks in
a single extent
```

(see #402 for the trace).

This change makes `recalculateBlocksize` actually pick a block size by
filesystem size, matching `mke2fs`'s small-vs-default split: 1 KiB
blocks below 512 MiB, 4 KiB blocks at or above. With 4 KiB blocks a
128 MiB journal needs exactly `maxBlocksPerExtent` (32768) blocks and
fits in a single extent.

Because `resize_inode` / reserved-GDT growth in go-diskfs currently
only supports 1 KiB blocks (there's an existing `TODO: Properly
support resize_inode` at the GDT-sizing block, and
`TestCreateWithBlockSizes` already passes
`WithFeatureReservedGDTBlocksForExpansion(false)` for non-1 KiB cases),
this also drops the feature when *we* picked the larger default. A
caller who explicitly sets `SectorsPerBlock` is unaffected: they pass
their own block size and the default-only path doesn't run.

## Test

Added `TestCreateMultiGB` — Create on 1 GiB and 2 GiB images, validated
with `e2fsck -f -n`. Both pass with the fix; the 2 GiB case is the
size that reproduces #402 on master. Full `go test ./...` is green.

## Behavior change to flag

This is a visible default change: existing callers who got 1 KiB
blocks back from `Create` on a >512 MiB filesystem will now get 4 KiB
blocks. Almost certainly the right default (matches `mke2fs`), but
worth calling out. Callers who need the old behavior can set
`Params.SectorsPerBlock = 2` explicitly.

## Not in this PR

Large-file writes that need >32768 blocks still hit the same inode
4-extent root limit fixed by no other change — `createRootExtentTree`
explicitly returns "cannot create root internal node" past 4 entries.
That's a separate, pre-existing limitation; this PR closes #402 for
the journal-allocation case without addressing it. Worth filing as its
own issue.

Fixes #402.

### DIFF
--- a/filesystem/ext4/ext4.go
+++ b/filesystem/ext4/ext4.go
@@ -328,8 +328,16 @@ func Create(b backend.Storage, size, start, sectorsize int64, p *Params) (*FileS
 
 	// recalculate if it was not user provided
 	if !userProvidedBlocksize {
-		sectorsPerBlockR, blocksizeR, numblocksR := recalculateBlocksize(numblocks, size)
+		sectorsPerBlockR, blocksizeR, numblocksR := recalculateBlocksize(size)
 		_, blocksize, numblocks = uint8(sectorsPerBlockR), blocksizeR, numblocksR
+		// resize_inode / reserved-GDT growth isn't yet supported for
+		// non-1 KiB block sizes (see TODO further down). When *we*
+		// picked the non-1 KiB default, drop the feature so Create
+		// doesn't try to lay out blocks it can't read back.
+		const oneKiB = uint32(SectorSize512) * 2
+		if blocksize != oneKiB {
+			fflags.reservedGDTBlocksForExpansion = false
+		}
 	}
 
 	// how many blocks in each block group (and therefore how many block groups)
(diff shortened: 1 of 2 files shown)


deitch (dismissed):

deitch:
This looks good to me; you even added the test. What it also needs is the `resize_inode` fix, but we do not have to hold it up. It is better to fix something broken and then make it even better, than to wait until we have everything. But your call. Claude probably would do a pretty good job filling in the missing pieces.

In any case, either mark it as ready and we can merge it in, or add the missing resize_inode piece and then mark it as ready.

Looks like [linting errors](https://github.com/diskfs/go-diskfs/actions/runs/26666525278/job/78804882773?pr=407)

deitch:
Something is off with the commit. It looks like it contains the expected ext4 fix and test, but it also reverts #404 in `squashfs/` and makes changes in `sync/` as well.

eriknordmark:
FWIW the ubuntu test failure is unrelated to these changes - a re-run should make it pass with the probability of 1/256 based on this analysis:

  Root cause — a pre-existing flaky test, not the ext4 change

  The test (fat32_test.go:600) is a negative case: corrupt the FAT32 FSInfo
  sector (which lives at byte offset 512, right after the boot sector) and
  assert that fat32.Read rejects it with "error reading FileSystem Information
  Sector".

  The bug is in how it corrupts (fat32_test.go:616-619):

  b := make([]byte, 1)
  _, _ = rand.Read(b)                 // one RANDOM byte
  _, _ = f.WriteAt(b, tt.bytechange+pre)   // written at offset 512

  Offset 512 is the first byte of the FSInfo sector's start signature —
  fsInfoSectorSignatureStart = 0x52526141 (fsinfosector.go:13), so that byte
  should be 0x52. Read only errors if signatureStart != 0x52526141
  (fsinfosector.go:47-48). The test writes a random byte there, so when
  rand.Read happens to produce 0x52, the byte is unchanged, the signature still
  validates, Read returns no error — and the test fails with actual <nil>
  expected error.

  That's a deterministic 1-in-256 flake baked into the test itself. The
  corruption isn't guaranteed to corrupt.

eriknordmark:
> Something is off with the commit. It looks like it contains the expected ext4 fix and test, but it also reverts #404 in `squashfs/` and makes changes in `sync/` as well.

I don't know what happened but it is cleaned up now.

deitch (approved):

deitch:
Looks good, merged in.

deitch:
OK, that is a nice catch on the test. I can fix that easily enough. Thank you.

PR #409, opened 2026-06-09
fix FSIS corruption test

deitch:
Thanks to @eriknordmark for finding it and commenting [here](https://github.com/diskfs/go-diskfs/pull/407#issuecomment-4650487324).

In short, in fat32 corruption tests, we replace a specific byte with a random one. Once in a while, that random one might be the same as the original, so nothing gets changed. This tests changes it instead to read the original one, and just add 10 and write it back.


deitch:
Ah, never mind. I did not realize @eriknordmark had fixed it in #408. I am closing in favour of that one.

PR #410, opened 2026-06-09
sync: skip excluded paths in CompareFS

eriknordmark:
## Problem

`CopyFileSystem` deliberately skips a small set of names — `lost+found`,
`.DS_Store`, `System Volume Information` — via its `excludedPaths` map, so
they are never written to the target. `CompareFS`, however, walked the
source and required *every* entry to also exist in the target, with no
such exclusion.

The two halves of the `sync` package therefore contradict each other:
verifying a faithful copy of any filesystem that contains a `lost+found`
(which every `mke2fs`-created ext4 does) fails with
`path "lost+found" missing in target FS`, even though the copy behaved
exactly as designed. The same latent failure exists for `.DS_Store` and
`System Volume Information`.

## Fix

Apply the same basename-based exclusion in both directions of `CompareFS`:
skip excluded entries (and their subtrees, via `fs.SkipDir`) when checking
that the source is present in the target, and again when scanning the
target for extra paths — so a target that has its own `lost+found` is not
reported as an extra path.

## Testing

`go test ./sync/...` passes. Added table cases to `TestCompareFS` covering
an excluded entry present only in the source, only in the target, and a
non-`lost+found` excluded name (`.DS_Store`); each fails before this change
with the exact `missing in target FS` / `extra path` signature and passes
after.

Verified end-to-end against the downstream consumer (partitionresizer): its
ext4 grow path creates a fresh target with `CreateFilesystem`, copies with
`CopyFileSystem`, then verifies with `CompareFS`; that verification
previously failed on `lost+found` and now succeeds.

### DIFF
--- a/sync/verify.go
+++ b/sync/verify.go
@@ -69,6 +69,15 @@ func CompareFS(origFS, targetFS fs.FS) error {
 		if err != nil {
 			return err
 		}
+		// Skip the same entries CopyFileSystem skips (lost+found, etc.):
+		// the copy never writes them to the target, so requiring them here
+		// would always fail. Match on basename, exactly as the copy does.
+		if excludedPaths[path.Base(p)] {
+			if d.IsDir() {
+				return fs.SkipDir
+			}
+			return nil
+		}
 		seen[p] = struct{}{}
 
 		// Check existence in target FS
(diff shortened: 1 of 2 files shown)


eriknordmark:
@deitch ready for review when you have a moment. This fixes an internal inconsistency in the `sync` package: `CopyFileSystem` skips `excludedPaths` (`lost+found`, `.DS_Store`, `System Volume Information`), but `CompareFS` required every source path in the target, so verifying a copy of any ext4 with a `lost+found` failed. `CompareFS` now applies the same exclusion in both walk directions. CI is green.

deitch (approved):

PR #411, opened 2026-06-17
backend,disk: make partition ops work on busy block devices

eriknordmark:
Two fixes needed to run partition tools (`e2fsck`/`resize2fs`) against the
partitions of a live block device. The file-based tests cannot reach either
path, so both were surfaced running a downstream resizer against a real
`/dev/nbd0`.

- **`file.OpenFromPathWithExclusive`** lets a caller open a device read-write
  *without* `O_EXCL` while still recording the path (`OpenFromPath` keeps its
  `O_EXCL` default). Holding the whole disk `O_EXCL` makes the kernel reject
  `O_EXCL` opens of its child partitions ("device is in use"), so a tool that
  shells out to `e2fsck`/`resize2fs` cannot hold the disk exclusively.

- **`ReReadPartitionTable`** falls back to per-partition **BLKPG**
  reconciliation when `BLKRRPART` fails — EBUSY during the udev re-probe a
  table write triggers, or a mounted sibling partition. BLKPG adds/removes/
  recreates only the changed entries via ioctls, with no whole-disk exclusive
  re-read and no dependency on external tools (`partx`).

On a regular file `O_EXCL` is a no-op and there are no partition nodes to
re-read, which is why the existing file-based tests never exercise either
path. Two root-gated loop-device tests cover the fixes:

- `disk/disk_blockdev_linux_test.go` exercises the BLKPG fallback: it mounts
  one partition to force `BLKRRPART` EBUSY, shrinks another, and verifies the
  kernel picks up the change while the mounted sibling is left untouched.

- `disk/disk_openexclusive_linux_test.go` exercises the non-exclusive open: a
  parent held `O_EXCL` blocks an `O_EXCL` open of its child partition (the
  `e2fsck`/`resize2fs` case) with EBUSY, while a non-exclusive parent leaves
  the child openable.

### DIFF
--- a/backend/file/file.go
+++ b/backend/file/file.go
@@ -28,6 +28,17 @@ func New(f fs.File, readOnly bool) backend.Storage {
 // Should pass a path to a block device e.g. /dev/sda or a path to a file /tmp/foo.img
 // The provided device/file must exist at the time you call OpenFromPath()
 func OpenFromPath(pathName string, readOnly bool) (backend.Storage, error) {
+	return OpenFromPathWithExclusive(pathName, readOnly, true)
+}
+
+// OpenFromPathWithExclusive is like OpenFromPath but lets the caller choose
+// whether a writable open uses O_EXCL. The exclusive flag is ignored for a
+// read-only open. Callers that shell out to filesystem utilities
+// (e.g. e2fsck/resize2fs) on the child partitions of a whole disk must open the
+// disk non-exclusively: holding the parent disk O_EXCL makes the kernel reject
+// O_EXCL opens of its partitions ("device is in use"). The returned backend
+// records the path, unlike New.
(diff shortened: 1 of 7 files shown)


eriknordmark:
@deitch could you take a look when you get a chance? This splits the two block-device fixes (non-exclusive whole-disk open + BLKPG re-read fallback) out of my downstream resizer work into go-diskfs, now with a root-gated loop-device regression test for the BLKPG path. Thanks!

deitch:
You have been pushing commits. Is this one ready to be reviewed?

eriknordmark:
> You have been pushing commits. Is this one ready to be reviewed?

Yes. No more commits planned.

deitch:
Overall, it looks good. Happy to have the extra test and handle the case. I am a little concerned that this is the first test that requires root to run, but it handles it with the skip correctly. So I am good with it.

deitch (approved):

PR #413, opened 2026-07-05
squashfs: handle images with no fragment or xattr table

maximilize:
## Problem

The squashfs reader fails on valid images written by [squashfs-tools-ng](https://github.com/AgentD/squashfs-tools-ng) that omit optional tables. Two independent failures, both reproducible with a minimal image (`gensquashfs -D <empty-dir> out.sqfs`):

- **No fragment table** (`fragmentCount == 0`): `readFragmentTable` reads unconditionally, but `fragmentTableStart` holds the `0xffffffffffffffff` "not present" sentinel, so `ReadAt` seeks to offset `-1` and fails with `error reading fragment table index: invalid seek position`.
- **No xattr table** (`NO_XATTRS` flag): `directoryEntryFromInode` calls `fs.xattrs.find(...)` when an inode advertises an xattr index, but `fs.xattrs` is `nil`, causing a **nil-pointer panic** in `(*xAttrTable).find` (`xattr.go:46`).

## Fix

Two guards, mirroring the "absent optional table" handling already done when the superblock is parsed (`!s.noXattrs && s.xattrTableStart != 0xffffffffffffffff`):

- skip reading the fragment table when `fragmentCount == 0`;
- skip the xattr lookup when `fs.xattrs == nil`.

## Tests

Added two internal regression tests (`TestReadFragmentTableNoFragments` and `TestDirectoryEntryFromInodeNilXattrTable`), both failing before / passing after. Full `filesystem/squashfs` suite passes and `go vet` is clean.

Verified end-to-end through rclone's `archive` backend (which uses this reader): the two images from [rclone/rclone#9004](https://github.com/rclone/rclone/issues/9004) that previously errored / panicked (a single empty directory, and a real squashfs-tools-ng Windows image) now list correctly.

### DIFF
--- a/filesystem/squashfs/directoryentry_internal_test.go
+++ b/filesystem/squashfs/directoryentry_internal_test.go
@@ -48,3 +48,25 @@ func TestDirectoryEntry(t *testing.T) {
 		t.Errorf("Expected InodeType %q for nil inode, got %q", "unknown", st.InodeType)
 	}
 }
+
+// TestDirectoryEntryFromInodeNilXattrTable checks that an inode advertising an
+// xattr index does not panic when the image has no xattr table. squashfs-tools-ng
+// sets the NO_XATTRS superblock flag (so fs.xattrs is nil) yet can leave a
+// non-sentinel xattr index in an inode. Regression test for rclone/rclone#9004.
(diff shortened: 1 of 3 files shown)


deitch (approved):
Solid catch. I feel like this shouldn't be possible, but it is, so catching it is correct.

PR #414, opened 2026-07-07
gpt: test round-trip of two partitions with the same name

eriknordmark:
A GPT partition is keyed by its unique GUID; the name field carries no
uniqueness constraint. A consumer that must select one of two
identically-named partitions therefore has to rely on the GUID surviving a
write/read round-trip. This adds a test that writes two `"EFI System"`
partitions with distinct GUIDs, reads the table back, and asserts the
labels stay identical while the per-entry GUIDs are preserved and distinct.

Test-only; no library change. Draft/WIP pending review — motivated by
downstream partition-selection-by-GUID work.

### DIFF
--- a/partition/gpt/table_duplicate_name_test.go
+++ b/partition/gpt/table_duplicate_name_test.go
@@ -1,0 +1,74 @@
+package gpt_test
+
+import (
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/diskfs/go-diskfs/partition/gpt"
+)
+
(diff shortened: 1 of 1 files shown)


eriknordmark:
@deitch this is ready for review — a small regression guard that two GPT partitions sharing a name but with distinct GUIDs round-trip through `Table.Write`/`Read` with per-entry GUIDs preserved. Groundwork for selecting a partition by GUID when labels are not unique. Thanks!

deitch (approved):

Issue #418, opened 2026-07-28
fat32: FAT is two entries too small at certain filesystem sizes


eriknordmark:
### Summary

For certain filesystem sizes, `fat32.Create` lays out a FAT that is exactly two entries too small for the cluster count it declares. The two highest clusters have no FAT entries, so the resulting filesystem is malformed and `fsck.vfat` rejects it:

```
Filesystem has 129024 clusters but only space for 129022 FAT entries.
```

The size has to be just right for this to happen, so most filesystems come out fine — but the affected sizes include round numbers such as 64 MiB and 4 GiB.

### Reproduction

```go
d, _ := diskfs.Create("fs.img", 64*1024*1024, 512)
d.CreateFilesystem(disk.FilesystemSpec{Partition: 0, FSType: filesystem.TypeFat32})
d.Close()
```

```
$ fsck.vfat -n fs.img
Filesystem has 129024 clusters but only space for 129022 FAT entries.
$ echo $?
1
```

Verified on `master` (c5b3b50) with `fsck.fat 4.2`. 63 MiB and 65 MiB are clean; 64 MiB is not.

### Root cause

`filesystem/fat32/fat32.go:131-136`:

```go
totalSectors := uint32(size / blocksize)
// Closed-form equivalent of the dosfstools mkfs.fat sectors-per-FAT search:
// smallest X such that (reserved + 2X + clusters*SPC) == totalSectors and
// X * (bytesPerSector/4) >= clusters + 2.
fatEntryDenom := uint32(blocksize)*uint32(sectorsPerCluster) + 8
sectorsPerFat := uint16((4*(totalSectors-uint32(reservedSectors)) + fatEntryDenom - 1) / fatEntryDenom)
```

Solving the two conditions stated in the comment for X gives

```
X = 4*(totalSectors - reservedSectors + 2*SPC) / (bytesPerSector*SPC + 8)
```

The numerator in the code omits the `+ 2*SPC` term, which is the contribution of the `+ 2` in `clusters + 2` — FAT entries 0 and 1 are reserved and cannot describe clusters. Usually the `+ fatEntryDenom - 1` rounding up absorbs the shortfall, but when the division comes out exact it does not, and the FAT ends up one sector short.

At 64 MiB with 512-byte sectors: `4*(131072-32)/520 = 1008.0` exactly, so `sectorsPerFat` is 1008, giving 129024 clusters and a FAT holding 129024 entries — of which only 129022 can name a cluster.

### Affected sizes

[...]

Issue #417, opened 2026-07-25
FAT: reading/copying a zero-length file fails — getClusterList rejects firstCluster == 0


eriknordmark:
### Summary

A zero-length file in a FAT12/16/32 filesystem legitimately has **no allocated clusters**, so its directory entry records a start cluster of `0`. When such a file is read (e.g. via `File.GetClusterChain` / a `Read`, or an `fs.Copy`/`sync.Copy` of the whole filesystem), go-diskfs errors out instead of treating it as an empty file:

```
unable to get list of clusters for file: invalid start cluster: 0
```

### Root cause

`filesystem/fat12/fat12.go`, `getClusterList`:

```go
func (fs *FileSystem) getClusterList(firstCluster uint32) ([]uint32, error) {
    if firstCluster > fs.table.MaxCluster() || fs.table.ClusterValue(firstCluster) == 0 {
        return nil, fmt.Errorf("invalid start cluster: %d", firstCluster)
    }
    ...
}
```

`firstCluster == 0` is not an error condition for a 0-byte file — it means the file has an empty cluster chain. `File.GetClusterChain` (`filesystem/fat12/file.go`) calls `getClusterList(fl.clusterLocation)` unconditionally and wraps the error as "unable to get list of clusters for file", so any read/copy of a 0-byte file fails.

### Impact

`github.com/diskfs/partitionresizer` copies a FAT filesystem when a partition moves (`resize.go`, via go-diskfs `sync.Copy`). If the source filesystem contains **any** zero-length file, the copy aborts. This was hit resizing an EFI System partition that contained a 0-byte marker file (`touch`-created), failing the whole partition move:

```
failed to copy FAT32 filesystem data for partition EFI System:
  copy dir boot: copy file boot/.boot_repository:
  unable to get list of clusters for file: invalid start cluster: 0
```

### Reproduction (sketch)

1. Create/open a FAT32 filesystem.
2. Create a zero-length file (e.g. `OpenFile(".../empty", O_CREATE|O_RDWR)` with no writes).
3. Read that file, call `GetClusterChain()` on it, or `sync.Copy` the filesystem → error.

### Suggested fix

A 0-byte file (directory-entry size 0, `firstCluster == 0`) should yield an **empty** cluster chain rather than an error. Either:

- short-circuit the file read/`GetClusterChain` path when the entry size is 0 (return an empty slice / EOF), or
- have `getClusterList(0)` return `([]uint32{}, nil)` for a legitimately empty file.

[...]

PR #419, opened 2026-07-28
fat32: fix zero-length files and dot-leading filenames

eriknordmark:
Up until now I've been successfully using partitionresizer to shrink and grow partitions used in Alpine Linux, but using live images. When I re-ran that using an installer image created using the mkimg.base in Alpine Linux there were two failures in go-diskfs:

1. Handling zero-length files in FAT32 created by Linux (issue #417).
2. Handling a filename like `.boot_repository` in a FAT32 filesystem.

Those both come from the zero-length `.boot_repository` created by `mkimg.base` — a plain `touch` in `scripts/mkimg.base.sh`, so the file is both zero length and dot-prefixed.

### Zero-length files with no allocated cluster

A file with no content has no cluster chain, and every FAT implementation other than go-diskfs records a start cluster of 0 for it. `getClusterList` rejected that outright, so reading such a file failed with `invalid start cluster: 0` and copying a filesystem aborted at the first zero-length file it reached. Reads now return EOF before looking up the chain, an empty file reports an empty cluster chain, and writing to one adopts the chain allocated for it.

go-diskfs also produces that representation itself now, rather than only tolerating it: a new file gets no cluster until it has content, and truncating a file to zero releases the clusters it held. `fsck.vfat` reports the previous behavior as `File size is 0 bytes, cluster chain length is > 0 bytes` and exits non-zero.

### Dot-leading filenames

A name beginning with a dot has nothing before the dot to build an 8.3 name from, and was stored with a blank one. `fsck.vfat` rejects those entries and renames them (`Bad short file name (.BOO)` → `FSCK0000.000`), and the directory listing skipped them, so the file was invisible to `Stat` and to any walk of the filesystem even though it was on disk — a file-level copy dropped it silently. Such a name now takes a stem derived from the rest of the name plus a numeric tail, matching mtools and the Linux kernel: `BOOT_R~1` for `.boot_repository`, with the existing collision handling picking `BOOT_R~2` for the next one. Listings also no longer drop an entry that has only a long name, so filesystems written by earlier releases stay readable.

One deliberate difference from mtools: for a name with a later dot, such as `.config.txt`, go-diskfs keeps the stem it already derives and produces `CONFIG.TXT` where mtools produces `CONFIG~1.TXT`. Both are valid and fsck accepts either; adding a tail unconditionally would change short names for many names that work today. The unit test records the difference.

### Tests

`sync/copy_fat32_test.go` copies a FAT32 partition holding zero-length files into a second partition, both file by file with `CopyFileSystem` and byte for byte with `CopyPartitionRaw`, and checks the target lists and reads back every entry. One case patches the source entry's start cluster to zero to get the representation other implementations write, without needing external tools.

`sync/copy_fat32_mtools_test.go` closes the loop outside go-diskfs: `mkfs.vfat` and `mcopy` create the source filesystem, go-diskfs copies the partition, then `mdir` and `fsck.vfat` inspect the result — so the filesystem under test is written and verified by something other than the code being tested. It skips when those tools are absent. Its partitions are 96 MiB because a 64 MiB FAT32 trips a separate layout bug (#418) that would make the fsck check fail for an unrelated reason.

There is also a unit test for the generated 8.3 names and their collision handling, and one asserting that a zero-length file holds no clusters whether it was created empty or truncated back to zero.

Each fix is a separate commit, and the tests come first, so they can be seen failing before either fix and passing after both: four failures at the test commit, two after the cluster fix, none after the short-name fix.


eriknordmark:
@deitch would you have time to look at this one?

I re-checked the two failures against current `master` (c5b3b50) before asking:

- The tests from the first commit, applied on top of `master` with no other changes, fail on all four cases — `copy file TEST.TXT: unable to get list of clusters for file: invalid start cluster: 0` for the zero-length files, and `.boot_repository` missing from the directory listing with `stat .boot_repository: file .boot_repository not found in directory .`
- The dot-leading name fails independently of file length: on `master`, a `.boot_repository` holding 22 bytes of content is created without error but is absent from `ReadDir`, `Stat` cannot find it, and a `CopyFileSystem` of that filesystem produces a target where the file does not exist at all. On this branch the same check lists it in both source and target and reads the content back.

The full suite passes on the branch (`go test ./...`, including ext4/squashfs/iso9660), and CI is green on the head commit — 20/20 checks, Build across all 17 GOOS/arch combinations plus Test on ubuntu, windows and macos.

One note on the external-tool test: `TestCopyFat32MtoolsRoundTrip` needs `mkfs.vfat`, `mcopy`, `mdir` and `fsck.vfat` and skips when they are not on PATH, so it exercises fully locally and, judging by the package runtime, skips on the runners. Happy to drop it or move it behind a build tag if you would rather the suite not reach for external tools.

deitch:
This looks pretty good. As far as I can tell, you always return empty clusters with zero-length files, but you also ensure that a 0-length file does not allocate a cluster list when created or written.

deitch (approved):

PR #420, opened 2026-08-01
ext4: fix wrong span for an internal node's last child

austinvazquez:
extentInternalNode.findBlocks and .blocks() called parseExtents with an inclusive end block in its count argument, but parseExtents' internal-node branch treats that argument as an actual count (start+count marks one past the end) when deriving the span of a node's last child -- real ext4 index entries don't store a count on disk, so the last entry's span has to come from the caller. The mismatch silently dropped the final file block of any range reaching the last child of a node loaded this way.

.blocks() reads were unaffected in practice: leaf-node parsing ignores the start/count arguments entirely, so a wrong count only mattered when it was also passed down to a child's own findBlocks call. findBlocks itself isn't wired up to any current caller, so this had no observable effect until something exercises it directly (e.g. a range read added on top of the existing full-file read path).

Pass the child's actual count instead of computing an end block, matching the one caller (loadChildNode) that already did this correctly.

### DIFF
--- a/filesystem/ext4/extent.go
+++ b/filesystem/ext4/extent.go
@@ -223,7 +223,7 @@ func (e extentInternalNode) findBlocks(start, count uint64, fs *FileSystem) ([]u
 		if err != nil {
 			return nil, err
 		}
-		ebf, err := parseExtents(b, e.blockSize, uint32(extentStart), uint32(extentEnd))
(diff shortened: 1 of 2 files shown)


deitch (approved):
Nice catch.

PR #421, opened 2026-08-01
ext4: stop discarding the cause of directory read failures

austinvazquez:
readDirWithMkdir's per-level readDirectory call dropped its error entirely, collapsing every possible cause -- checksum mismatches, unparseable dirents, htree rejections, extent tree failures, inode read failures -- into a single opaque "failed to read directory <path>" with no detail at all. The mkSubdir wrapper a few lines above had the same problem. The root-level readDirectory call did include the error text (%v), but not in a form errors.Is/errors.As can reach.

Wrap all three with %w so the real cause survives to the caller.


deitch (approved):
This seems rather obvious in retrospect, doesn't it?

PR #423, opened 2026-08-03
partition: add GPT and MBR selection policies

### LINKED ISSUE #422: partition.Read selects stale GPT over active MBR — GPT result is returned when ProtectiveMBR is false
### Bug

`partition.Read()` always tries GPT first and immediately returns the GPT table
when `gpt.Read()` succeeds.

```go
// partition/partition.go L15-23
gptTable, err := gpt.Read(f, logicalBlocksize, physicalBlocksize)
if err == nil {
	return gptTable, nil
}

mbrTable, err := mbr.Read(f, logicalBlocksize, physicalBlocksize)
if err == nil {
	return mbrTable, nil
}
```

This produces a different result from Linux when a disk contains:

- a valid conventional MBR at LBA 0;
- no `0xEE` protective-MBR entry;
- a stale but CRC-valid GPT header and partition-entry array at LBA 1.

`gpt.Read()` successfully parses the stale GPT and sets
`gptTable.ProtectiveMBR` to `false`, but `partition.Read()` does not check that
field and returns the GPT table without trying MBR.

Source:

- [`partition.Read`](https://github.com/diskfs/go-diskfs/blob/master/partition/partition.go)
- [`readProtectiveMBR`](https://github.com/diskfs/go-diskfs/blob/master/partition/gpt/table.go)

Linux also probes GPT before MBR, but its GPT parser normally requires a valid
protective or hybrid MBR containing a `0xEE` entry. If that check fails, Linux
continues with the MBR parser.

Linux source:

- [`block/partitions/core.c`](https://github.com/torvalds/linux/blob/master/block/partitions/core.c)
- [`block/partitions/efi.c`](https://github.com/torvalds/linux/blob/master/block/partitions/efi.c)

### Example

The affected disk contains both DOS and GPT signatures:

```text
$ blkid -p /dev/vdb
/dev/vdb: PTUUID="8a6e23ea" PTTYPE="dos"
```

```text
$ wipefs -n /dev/vdb
DEVICE OFFSET TYPE UUID LABEL
vdb    0x1fe  dos
vdb    0x200  gpt
```

Linux selects the MBR and creates MBR-style PARTUUID links:

```text
$ ls -l /dev/disk/by-partuuid/
8a6e23ea-01 -> ../../vdb1
8a6e23ea-02 -> ../../vdb2
8a6e23ea-03 -> ../../vdb3
```

However, `partition.Read()` returns `*gpt.Table`, and the partitions contain
GPT GUIDs such as:

```text
DD1A39D1-459E-8F4E-B7E1-2B8C0ABDE4FD
D34CBE52-6E38-3141-BB64-B780E5FB4BEA
C6FB24A5-A02C-3C4D-9CC3-7B184F79BCD1
```

| Component | Selected table | Partition identifier |
| --- | --- | --- |
| go-diskfs `partition.Read()` | stale GPT | GPT partition GUID |
| Linux kernel / udev | active MBR | `8a6e23ea-01` |
| `blkid` / `fdisk` | active MBR | DOS disk identifier |

The GPT and MBR partition entries can describe similar sector ranges, so the
wrong selection may not be obvious until the returned partition UUID is used.

### Impact

Any application that uses `partition.Read()` and then resolves the returned
partition UUID through:

```text
/dev/disk/by-partuuid/<partition-uuid>
```

can fail on these disks.

In our case, a read-only disk scanner received GPT partition GUIDs from
go-diskfs, while Linux/udev only created MBR-style PARTUUID links. The expected
device paths never appeared, so the partitions could not be mounted or scanned.

Retries and `partprobe` do not fix the failure because this is not udev timing:
go-diskfs and the Linux kernel selected different partition tables.

### Proposed Direction

I do not think `gpt.Read()` should stop parsing GPT data without a protective
MBR. Reading that data is useful for recovery and forensic use cases.

The issue is the authoritative result returned by the generic
`partition.Read()` auto-detection API.

Would the maintainers accept an opt-in Linux-compatible detection mode, for
example:

```go
partition.ReadWithOptions(file, logicalBlockSize, physicalBlockSize,
	partition.ReadOptions{
		Policy: partition.LinuxCompatible,
	},
)
```

The Linux-compatible policy could:

1. parse and validate the GPT header and partition-entry array;
2. classify LBA 0 as invalid, protective, or hybrid MBR;
3. select GPT only when a protective/hybrid MBR is present, unless explicitly
   forced;
4. otherwise continue with MBR detection.

A direct check of the current `ProtectiveMBR` boolean may be too strict because
`readProtectiveMBR()` requires a pure protective MBR and may reject legitimate
hybrid-MBR layouts or size mismatches after a disk image is copied to a larger
device.

I can provide a minimized regression image and prepare a PR after the expected
API and compatibility behavior are agreed.

### Environment

- go-diskfs `v1.9.2-0.20260429091007-50be90f6cfae`
- issue reproduced in an application using that version
- selection logic confirmed unchanged in go-diskfs `v1.9.4`
- Linux, 512-byte logical sectors

wow20220809-lab:
## Summary

`partition.Read` currently returns GPT as soon as GPT can be read. This can
select readable residual GPT metadata even when the disk contains a valid
conventional MBR.

This change introduces two explicit selection policies:

- `GPTWithMBR`: prefers a valid conventional MBR while continuing to use GPT
  for protective and hybrid MBR layouts. This is the new default.
- `GPTIgnoreMBR`: preserves the previous GPT-first behavior.

The existing `partition.Read` signature remains unchanged and delegates to
`ReadWithPolicy` using `GPTWithMBR`.

## Behavior

| GPT | MBR layout | GPTWithMBR | GPTIgnoreMBR |
| --- | --- | --- | --- |
| Valid | No valid MBR | GPT | GPT |
| Valid | Protective MBR | GPT | GPT |
| Valid | Hybrid MBR | GPT | GPT |
| Valid | Conventional MBR without valid 0xEE entry | MBR | GPT |
| Invalid | Valid MBR | MBR | MBR |
| Invalid | Invalid MBR | Error | Error |

A protective entry is recognized by type `0xEE` with starting LBA 1. Its size
is intentionally not required to span the current disk, since a disk image may
have been copied to a larger device without updating the PMBR size.

## Tests

Added coverage for:

- default `GPTWithMBR` behavior
- conventional MBR with readable residual GPT metadata
- `GPTIgnoreMBR`
- protective and hybrid MBR layouts
- missing or invalid MBR
- invalid protective entries
- invalid GPT fallback
- invalid policy values

Validation performed:

- `go test ./partition/... -count=1`
- `go test -race ./partition/... -count=1`
- `go vet ./...`
- `golangci-lint run ./partition/...`

Fixes #422


deitch (approved):
This is great, thank you.

deitch:
Thanks, merged in. Who are you @wow20220809-lab ; and how do you use it?

wow220809:
Thanks for merging it!

I'm a cloud security engineer. Detecting security risks in cloud disks is an important part of our work, and go-diskfs helps us identify disk layouts and mount those disks for comprehensive security scanning.

Thank you very much for creating and maintaining this project!

PR #424, opened 2026-08-03
ext4: accept non-zero padding bytes in HTree dot/dotdot names

austinvazquez:
parseDirectoryTreeRoot validated the full 4-byte name fields of the "." and ".." directory entries against ['.', 0, 0, 0] / ['.', '.', 0, 0]. The ext4 spec only guarantees that the first name_len bytes are meaningful; the trailing padding bytes may be non-zero and the Linux kernel does not check them.

The strict check broke ReadDir on any HTree-indexed directory whose block happened to contain non-zero padding, e.g. a large __pycache__ directory where bytes 0xa–0xb of the dot entry were observed as '/' (0x2f) and 'a' (0x61).

Fix both checks to compare only the first name_len bytes and remove the now-unused bytes import.

### DIFF
--- a/filesystem/ext4/directory.go
+++ b/filesystem/ext4/directory.go
@@ -144,7 +144,7 @@ func parseDirectoryTreeRoot(b []byte, largeDir bool) (node *directoryHashRoot, e
 	// Only check the meaningful name bytes (name_len=1); padding bytes beyond
 	// that are not guaranteed to be zero by the ext4 spec.
 	if b[0x8] != '.' {
-		return nil, fmt.Errorf("directory hash tree root dot name is %s and not '.'", b[0x8:0x9])
+		return nil, fmt.Errorf("directory hash tree root dot name is %q and not '.'", b[0x8:0x9])
 	}
 
 	// dotdot parameters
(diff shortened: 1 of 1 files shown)


austinvazquez:
For this one, I have a use case which is failing to parse a pycache directory where the bytes 0xa-0xb of the dot entry are '/' and 'a' instead of zero. Reading the on-disk format [spec](kernel.org/doc/html/latest/filesystems/ext4/directory.html), it certainly seems these should be zero, but even the kernel code ([ref](https://github.com/torvalds/linux/blob/848acc8ffe1b7cd5f1bf427b93069becfebc2c9d/fs/ext4/namei.c#L1405)) itself only checks the `name_len` bytes from the field and not reading beyond that.

So perhaps it is ok to only be as strict as the kernel here.

austinvazquez:
A review bot recommended https://github.com/diskfs/go-diskfs/pull/424/commits/96d6229afb33bfa4116e8b5088d33858f8a67401 to me. Pushed here as separate commit while I have the hood open.

deitch (approved):
This one really is funny. But if the kernel is not strict about it, no reason we should be.

deitch:
👋 @austinvazquez 

You're a Docker and containerd maintainer? We probably have lots of common friends. I will pop you a LinkedIn, we can say hello.

PR #426, opened 2026-08-10
ext4: fix bitmap indexes and metadata cleanup in Remove()

### LINKED ISSUE #425: ext4: Remove() corrupts data via wrong block/inode bitmap indexes
### Summary
`FileSystem.Remove()` in the ext4 implementation has two off-by-one bugs in bitmap index calculation that cause silent data corruption.

### Problem

When using go-diskfs to manage ext4 images, removing a file and then creating a new one can **silently corrupt other files** on the same image. In practice this is easily triggered by editors that save via a temp-file-then-rename pattern (e.g. Vim swap file workflow):

1. Editor creates a temp file and writes new content to it.
2. Editor deletes the original file.
3. Editor renames the temp file to the original name.

After step 2–3, the original file's content is lost or replaced by another file's content. `e2fsck -f -n` reports serious errors:

multiply-claimed block(s)
Inode bitmap differences
Unattached inode
Inode N ref count is X, should be Y
Free blocks count wrong

### Reproduction

Minimal reproducer with a 1 KiB-block image (the default for sizes < 512 MB):

1. Create `/a` (4096 bytes of `A`), `/b` (`B`), `/c` (`C`)
2. Remove `/b`
3. Create `/d` (`D`)

**Expected:** `/a`, `/c`, `/d` all intact.
**Actual:** `/c` is corrupted — reads as `DDDD…` because `/d` reused `/c`'s
inode. `e2fsck` reports `multiply-claimed block(s)` between `/c` and `/d`.

### Root Causes

`FileSystem.Remove()` in `filesystem/ext4/ext4.go` has five defects:

#### 1. Block bitmap index ignores `firstDataBlock`

```go
blockInBG := int(i) - int(fs.superblock.blocksPerGroup)*bg
```
On 1 KiB-block images firstDataBlock = 1, so the group's first block is firstDataBlock + bg*blocksPerGroup. Dropping firstDataBlock clears the adjacent block's bit — the target block stays allocated and an in-use block is freed instead, causing two files to share one physical block.

#### 2. Inode bitmap index missing -1
```go
inodeInBG := int(entry.inode) - int(fs.superblock.inodesPerGroup)*inodeBG
```
Inode numbers are 1-based; the bitmap is 0-based. Without -1, removing inode N clears the bit for inode N+1, so the next Create reuses that inode while it is still referenced by an existing file.

#### 3. Parent directory hardLinks not decremented on subdir removal
Removing an empty directory decrements gd.usedDirectories but never decrements the parent inode's hardLinks. e2fsck reports the mismatch.

#### 4. Removed inode's table entry is not wiped
The current code deliberately skips wiping the inode table entry, as noted in a comment at the end of `Remove()`:
```go
// we could remove the inode from the inode table in the group descriptor,
// but we do not need to do so. Since we are not reusing the inode, we can
// just leave it there, the bitmap always is checked before reusing an
// inode location.
```
Only the bitmap is cleared; the inode table keeps hardLinks, size, blocks, and extents from the deleted file. If e2fsck -y is run, it resurrects the stale inode into lost+found and re-claims data blocks that may have already been reallocated to new files.
> The fix wipes hardLinks, size, blocks and sets deletionTime so the image passes e2fsck -f -n cleanly. If the maintainer considers this out of scope, this is the one fix that can be dropped without causing data loss in go-diskfs's own operations

#### 5. Free block count double-counted and wrong unit
removedInode.blocks (in 512-byte sectors) is added to both the group descriptor and the superblock free counters, on top of the per-group freedByBG accumulation that already ran. This double-counts and mixes sector units with filesystem-block units.

### Environment
- go-diskfs: master (d43e027)
- Block size: 1 KiB (default for images < 512 MB)

Morolis:
- block bitmap: subtract firstDataBlock so the correct bit is cleared on 1 KiB-block images
- inode bitmap: add -1 offset for 1-based inode numbers
- decrement parent directory hardLinks when removing a subdirectory
- wipe the removed inode table entry to prevent e2fsck -y resurrecting stale inodes and reclaiming already-reallocated blocks
- fix free block counter to use actually freed filesystem blocks instead of double-counting via removedInode.blocks

Add TestRemoveIntegrity covering all five regressions.

Fixes #425

### DIFF
--- a/filesystem/ext4/ext4.go
+++ b/filesystem/ext4/ext4.go
@@ -1436,7 +1436,10 @@ func (fs *FileSystem) Remove(p string) error {
 				blockBitmaps[bg] = dataBlockBitmap
 			}
 			// the extent lists the absolute block number, but the bitmap is relative to the block group
-			blockInBG := int(i) - int(fs.superblock.blocksPerGroup)*bg
+			// group bg's first block is firstDataBlock + bg*blocksPerGroup (firstDataBlock
+			// is 1 on 1 KiB-block images); ignoring firstDataBlock clears the wrong bit.
+			groupStart := uint64(fs.superblock.firstDataBlock) + uint64(bg)*uint64(fs.superblock.blocksPerGroup)
+			blockInBG := int(i - groupStart)
 			if err := dataBlockBitmap.Clear(blockInBG); err != nil {
 				return fmt.Errorf("could not clear block bitmap for block %d: %v", i, err)
 			}
@@ -1519,7 +1522,8 @@ func (fs *FileSystem) Remove(p string) error {
 
 	// remove the inode from the bitmap and write the inode bitmap back
 	// inode is absolute, but bitmap is relative to block group
-	inodeInBG := int(entry.inode) - int(fs.superblock.inodesPerGroup)*inodeBG
+	// bitmap start at 0, inode numbers start at 1, so need off-by-one
+	inodeInBG := int(entry.inode) - int(fs.superblock.inodesPerGroup)*inodeBG - 1
 	if err := inodeBitmap.Clear(inodeInBG); err != nil {
 		return fmt.Errorf("could not clear inode bitmap for inode %d: %v", entry.inode, err)
 	}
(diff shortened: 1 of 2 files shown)


deitch (approved):
Thank you again for this.

PR #427, opened 2026-08-12
fat32: fix FAT sizing for reserved entries and large volumes

### LINKED ISSUE #418: fat32: FAT is two entries too small at certain filesystem sizes

eriknordmark:
Fixes #418.

Two FAT sizing bugs in `fat32.Create`, one per commit.

**Reserved FAT entries.** The sectors-per-FAT calculation omitted the term for entries 0 and 1, which are reserved and name no cluster. Rounding up hid it except where the division came out exact, leaving the two highest clusters with no FAT entry at sizes such as 64 MiB, 129 MiB, 259 MiB and 4 GiB. `fsck.vfat` rejects those: `Filesystem has 129024 clusters but only space for 129022 FAT entries`.

**Truncation above 256 GiB, overflow above 512 GiB.** Sectors-per-FAT was computed into a `uint16` although the boot sector field is 32-bit, so 257 GiB at 512-byte sectors stores 240 instead of 65776. Above 512 GiB the intermediate `4*(totalSectors-reserved) + 8*sectorsPerCluster` also exceeds 32 bits, the quotient comes out zero, and `Create` panics allocating the FAT (`index out of range [2] with length 1`) — at sizes `Fat32MaxSize` accepts. Both limits are reachable only with 512-byte sectors.

The sizing moved into a `layout()` helper so the invariant can be swept rather than sampled: every MiB from 16 MiB to 4 GiB and every GiB up to `Fat32MaxSize`, at both sector sizes, must satisfy `FAT entries >= clusters + 2`. Reverting the first fix fails 48 of the 8162 MiB-range combinations; reverting the second fails every size from 257 GiB up.

Verified against `fsck.vfat 4.2`: 64/129/259/4096 MiB and 512 GiB/1 TiB images come out clean. Layouts move only where they must — across a 1166-size sample, 9 change: the 6 that fsck rejected, plus 3 whose FAT had exactly zero spare entries and now gains one sector.

@deitch


deitch:
I am confused by the second commit, i.e. making `sectorsPerFat` uint64. In the FAT32 EBPB (see the [Wikipedia entry](https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system#FAT32_Extended_BIOS_Parameter_Block) or [osdev doc](https://wiki.osdev.org/FAT#FAT_32_2), sectors per FAT is 4 bytes. This further aligns with when it is marshaled/unmarshaled into the structure in [FromBytes](https://github.com/diskfs/go-diskfs/blob/master/filesystem/fat32/dos71bpb.go#L95) and [toBytes](https://github.com/diskfs/go-diskfs/blob/master/filesystem/fat32/dos71bpb.go#L156).

How do we just make it a uint64? 

Or are you saying to keep it a 32, but that the filesystem size as a whole needs to be 64?

eriknordmark:
Sectors-per-FAT stays 32-bit — the EBPB field and `dos71EBPB.sectorsPerFat` are
untouched. The bug was that `layout`'s local was `uint16`, so a 32-bit field was
being fed from a 16-bit value: at 512-byte sectors the correct value passes 65535
at 257 GiB (256 GiB gives 65520, just under) and truncates — 257 GiB stores 240.

Separately, the intermediate `4*(totalSectors-reserved) + 8*sectorsPerCluster`
needs 34 bits at `Fat32MaxSize`, even though both the input (`totalSectors`) and
the result (`sectorsPerFat`) are 32-bit. Only those temporaries become `uint64`.
At 512 GiB it wraps to 32647, the quotient becomes 0, and `Create` panics on
`clusters[rootDirCluster]`.

So: the field stays 32, the arithmetic in between needs 64, and the size in bytes
is already `int64`. Both effects are 512-byte-sector only — at 4096 the largest
sectors-per-FAT in range is 65512, which still fits `uint16`.

My commit message fused the two thresholds ("above roughly 256 GiB ... overflowed
32 bits"); I'll reword it to name them separately.

deitch:
Ah, got it, now it makes sense. Thanks for clarifying.

deitch (approved):

PR #428, opened 2026-08-16
ext4: validate directory entry record and name lengths

hibrian827:
## What

Parsing an untrusted ext4 image can panic when a directory entry has an inconsistent length field.

`directoryEntryFromBytes` sliced the name with the on-disk `name_len` without checking it fits the record, and `parseDirEntriesLinear` sliced each entry with the on-disk `rec_len` without validating it. A crafted directory block (e.g. a 12-byte entry with `name_len=250`) reaches these slices via the public `ext4.Read(...).ReadDir(...)` path and panics with `slice bounds out of range` instead of returning an error.

## Fix

- `directoryEntryFromBytes`: compute the name end with `int` arithmetic  and return an error if the name extends past the record.
- `parseDirEntriesLinear`: return an error for truncated headers and for record lengths below the minimum or beyond the remaining buffer.

Added regression tests for both.


deitch (approved):
Solid. And simple, as the best ones are. Thank you for this.

PR #430, opened 2026-08-19
ext4: fix extent skip condition in Read/Write causing panic

### LINKED ISSUE #429: ext4: File.Read/Write panic (makeslice: len out of range) on fragmented files at unaligned offsets
### Summary
`File.Read()` and `File.Write()` panic with `runtime error: makeslice: len out of range` when a file has **multiple (fragmented) extents** and the file offset lands inside a later extent at a **non-block-aligned** position.

### Problem
When using go-diskfs to manage ext4 images,  and renaming a file larger than 4096 using my own implementation, pain occurs. My Rename is simple. its rename files by copying them. When the file larger than 4006(the standard size of extent), painc:  `runtime error: makeslice: len out of range` 

detailed logs:
panic: runtime error: makeslice: len out of range
goroutine 1 running:
github.com/diskfs/go-diskfs/filesystem/ext4.(*File).Read(...)
        filesystem/ext4/file.go:73

My Rename Step:
1. stat target files and source files
2. open target files by creating and truncating mode 
3. copy data from source files -> painc
4. remove source file

### Root Cause
Both `Read()` and `Write()` iterate over the file's extents and skip extents that end before the current offset's block:

```go
readStartBlock := uint64(fl.offset) / blocksize
for _, e := range fl.extents {
    if uint64(e.fileBlock)+uint64(e.count) < readStartBlock {   // BUG: should be <=
        continue
    }
    extentSize := int64(e.count) * int64(blocksize)
    startPositionInExtent := fl.offset - int64(e.fileBlock)*int64(blocksize)
    leftInExtent := extentSize - startPositionInExtent          // goes NEGATIVE
    ...
    b2 := make([]byte, toReadInOffset)                          // panics
```
e.fileBlock + e.count is the first file block after the extent(half-open range). An extent covering file blocks [0, 2) ends exactly at block 2: if readStartBlock == 2, the extent is fully behind the read window and must be skipped — but < fails to skip it. For a non-aligned offset the stale extent then computes:
- startPositionInExtent = offset - fileBlock*blocksize — larger than the extent size
- leftInExtent = extentSize - startPositionInExtent — negative
- make([]byte, negative) → panic

With block-aligned offsets the skip accidentally works(readStartBlock == fileBlock+count implies startPositionInExtent == extentSize, leftInExtent == 0), which is why this only bites fragmented files read at unaligned offsets.

### Reproducer
use 1 KiB-block image
```go
fs, _ := ext4.Create(file.New(f, false), 100*MB, 0, 512, &ext4.Params{})

// /t.dat: extent A = fileBlock 0 count 2, then pad files grab blocks,
// then extent B = fileBlock 2 count 2  (fragmented)
w, _ := fs.OpenFile("/t.dat", os.O_CREATE|os.O_RDWR)
w.Write(bytes.Repeat([]byte("C"), 2048))
w.Close()
for i := 0; i < 3; i++ {
    p, _ := fs.OpenFile(fmt.Sprintf("/pad%d.dat", i), os.O_CREATE|os.O_RDWR)
    p.Write(bytes.Repeat([]byte("Q"), 1024))
    p.Close()
}
w2, _ := fs.OpenFile("/t.dat", os.O_RDWR)
w2.Seek(2048, 0)
w2.Write(bytes.Repeat([]byte("D"), 2048))
w2.Close()

// read at unaligned offset inside extent B
r, _ := fs.OpenFile("/t.dat", os.O_RDONLY)
r.Seek(3000, io.SeekStart)
n, err := r.Read(make([]byte, 100))   // panics on master
```
Both Read (file.go:73) and Write (file.go:198) are affected.

### Environment
- go-diskfs master (26c6e3a)
- any block size; easiest reproduced with 1 KiB blocks

Morolis:
An extent covers file blocks [fileBlock, fileBlock+count). The skip condition used a strict <, so an extent ending exactly at the current read/write start block was not skipped. With a non-block-aligned offset this made leftInExtent negative and make([]byte, n) panicked with "makeslice: len out of range".
Use <= in both Read and Write

Fixes #429

### DIFF
--- a/filesystem/ext4/file.go
+++ b/filesystem/ext4/file.go
@@ -55,13 +55,20 @@ func (fl *File) Read(b []byte) (int, error) {
 	readStartBlock := uint64(fl.offset) / blocksize
 	for _, e := range fl.extents {
 		// if the last block of the extent is before the first block we want to read, skip it
-		if uint64(e.fileBlock)+uint64(e.count) < readStartBlock {
+		// notice: need equal sign
+		if uint64(e.fileBlock)+uint64(e.count) <= readStartBlock {
 			continue
 		}
 		// extentSize is the number of bytes on the disk for the extent
 		extentSize := int64(e.count) * int64(blocksize)
 		// where do we start and end in the extent?
 		startPositionInExtent := fl.offset - int64(e.fileBlock)*int64(blocksize)
+
+		// Protect
+		if startPositionInExtent < 0 || startPositionInExtent > extentSize {
+			return int(readBytes), fmt.Errorf("read offset %d falls outside extent (fileBlock %d, count %d) of file %s", fl.offset, e.fileBlock, e.count, fl.filename)
+		}
+
 		leftInExtent := extentSize - startPositionInExtent
 		// how many bytes are left to read
 		toReadInOffset := bytesToRead - readBytes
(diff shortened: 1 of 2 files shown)


deitch (approved):
And you used the reproduction from the issue you opened in the test; thank you for doing so.

PR #433, opened 2026-09-14
Return an error when an El Torito parent directory is missing

youdie006:
`findEntry` returns `(nil, nil)` when nothing matches (`filesystem/iso9660/finalize.go:498`), and `Finalize` dereferences the `parent` it gets back without checking — at `:680`, `:690` and `:692`. The `child` from the identical call three lines later **is** nil-checked, at `:697`, in the same `if/else` block.

So an `ElTorito.BootCatalog` or `BootFile` whose directory is not in the workspace panics instead of returning an error:

```
iso9660.(*finalizeFileInfo).findEntry(0x0, {0x6a4c1f, 0x9})   finalize.go:480
iso9660.(*FileSystem).Finalize(...)                            finalize.go:692
```

The two added guards mirror the wording and shape of the existing `child == nil` check rather than introducing a new style.

<details>
<summary>Verification</summary>

Base `5ec1662`. Two subtests, one per field, both driving the public API (`Create` → `OpenFile` → `Finalize`) with no stubs.

| row | `finalize.go` md5 | subtests that ran | result |
|---|---|---|---|
| fix | `cc1ad0e0…` | 2 | both pass |
| base | `1b7ef8f8…` | 1 | **panic** |
| catalog guard removed only | `fafff937…` | 1 | `boot_catalog_parent_missing` fails |
| image guard removed only | `672517d5…` | 2 | catalog **passes**, `boot_image_parent_missing` fails |

The last row is the one that matters: with only the image guard removed the first subtest still passes, so each guard is pinned by its own subtest and neither is a passenger. On the base and catalog-removed rows only one subtest runs because the panic takes the test binary down — that is the defect, not a filter miss.

`go test -count=1 ./...` — all 14 packages ok. Gates as `.github/workflows/ci.yaml` writes them: `go build` ok, cross-compile `GOOS=windows/amd64` and `GOOS=darwin/arm64` ok, `make vet` exit 0, `gofmt -l` clean on both changed files. `golangci-lint run ./...` reports the same finding set before and after, none at the changed lines — though my linter is 2.13.2 while the Makefile pins v2.5.0, so treat that row as indicative. I did not run `make test`, which needs a Docker build.

</details>

While looking at this I also checked whether `Open` accepting a backslash path was a defect, and it is not: `filesystem/iso9660/iso9660_internal_test.go:47-50` and `:105-106` assert backslash-as-separator on the read path, so `universalizePath` is deliberate. Nothing in this PR touches it.

---

Disclosure: written with the help of Claude (an AI assistant). Every number above is from runs on my machine.


deitch:
In theory, [`findEntry()`](https://github.com/diskfs/go-diskfs/pull/433?email_source=notifications&email_token=AACE7HWFZNLNSUTQ7RYE4DD5O6WB3A5CNFSNUABEM5UWIORPF5TWS5BNNB2WEL2QOVWGYUTFOF2WK43UF42DKMRUGU3DCNJWHGTHEZLBONXW5KTTOVRHGY3SNFRGKZFFMV3GK3TUVRTG633UMVZF6Y3MNFRWW) is supposed to return an error if it could not find the entry. That having been said, it certainly does not hurt to add the safety check, and it looks like your tests manage to cause it to happen?

deitch (approved):

PR #434, opened 2026-09-16
iso9660: treat an interior backslash as a name character

youdie006:
`Open` splits a path with `path.Dir`/`path.Base` (`filesystem/iso9660/iso9660.go:473-474`), which recognise `/` only, then hands the directory half to `readDirectory`, where `splitPath` runs `universalizePath` and rewrites every backslash to a slash (`filesystem/iso9660/util.go:18-25`). The two halves of one path end up split by different rules:

```
Open("DEEP/A/B")   -> opens DEEP/A/B
Open("DEEP\A/B")   -> opens DEEP/A/B      path.Dir gives "DEEP\A", readDirectory rewrites it
Open("DEEP\A")     -> file does not exist  path.Base is matched literally
Open("DEEP\A\B")   -> file does not exist
```

`validatePath` already defers to `iofs.ValidPath` (`iso9660.go:753-755`), and io/fs defines the backslash as an ordinary name character, so `fs.ValidPath("DEEP\\A/B")` is true and the path must not resolve. Anything that hands a user-supplied name to the io/fs surface reaches this: `http.FileServer(http.FS(fsys))`, `fs.WalkDir`, `fs.Sub`. The other direction is also wrong: an entry whose Rock Ridge or Joliet name contains a backslash cannot be opened.

The rewrite is kept for a leading backslash, which is the Windows absolute form that `readDirectory`'s own tests pin (`iso9660_internal_test.go:47`, `:49`, `:105`); `\FOO\BAR` still becomes `/FOO/BAR`. A path that does not start with one is left alone.

<details>
<summary>Verification</summary>

`go test -count=1 ./filesystem/iso9660/`. `md5sum filesystem/iso9660/util.go` on every row.

| `util.go` | md5 | result |
| --- | --- | --- |
| this PR | `eb8c1576` | `ok ... 0.399s` (whole package) |
| pristine `HEAD` | `efffcdb8` | 3 failures: `Open`, `Stat`, `ReadDir` on the backslash forms |
| drop `universalizePath` from `splitPath` entirely | `e9b590fa` | 3 **pre-existing** failures: `readDirectory(\)` and `readDirectory(\FOO)` at `iso9660_internal_test.go:56`, `readDirectory(\)` at `:112` |

The two directions fail disjoint sets: the revert kills only the new assertions, the blanket removal kills only rows that were already there. The new test also asserts the controls, that `Open("DEEP/A/B")` and `ReadDir("DEEP/A")` still succeed, and that `fs.ValidPath` is true for each backslash form so the test keeps testing what it means to.

`go test ./... -count=1` 14 packages ok, `go vet ./filesystem/...` clean. `gofmt -l filesystem/iso9660/` lists only `statt_windows.go`, which is pre-existing and untouched here. No new testdata: `DEEP/A/B` already ships in `9660.iso`.

Two other callers of `splitPath` take the same rewrite and are not covered by this test: `rockridge.go:530` splits a symlink target for the SL record, and `finalize.go:474` is on the ISO-creation path.

</details>

One thing worth naming: `fat12` and `fat32` do the same rewrite in their own `universalizePath`, so this is not an isolated outlier. I left them alone because a backslash is a real separator on those filesystems, which is a different question from what io/fs says about a name. Happy to look at them if you would rather they match.

Disclosure: this change was written with AI assistance (Claude). The measurements above were reproduced and checked by me before opening this PR.


deitch:
I am trying to wrap my head around this and am confused. Right now, `universalizePath(p)` converts all `\` to `/`, essentially saying, `\` is a separator, not a character. So `A\B\C` and `A/B/C` are treated identically as `A -> B -> C` (walking the tree).

You are suggesting that we do that only if the path is absolute and starts with `\`. 

* `A/B/C` means `A -> B -> C` (C child of B child of A)
* `\A\B\C` means `A -> B -> C` (C child of B child of A)
* `A\B\C` means `A\B\C` (single file in root `A\B\C` with 5 chars in its name)

Is that correct? Why would we want `\` to be a valid separator only if it starts with `\` ie absolute path? Is there a need to support `\` as an ordinary filename character in iso9660? It is not in [the iso9660 spec](https://ecma-international.org/wp-content/uploads/ECMA-119_2nd_edition_december_1987.pdf) (see 7.4 and 7.5), where it is solely d-characters or d1-characters.

Issue #310, opened 2025-11-23
Cannot read from Raspberry Pi OS partition.


StoneMonarch:
When trying to read from a Raspberry Pi OS image I get a `divide by zero` error on the ext4 partition.

Reproduce:

1. Download any official image from the Raspberry Pi [website](https://www.raspberrypi.com/software/operating-systems/)
2. Extract and try to read the partitions. FAT32 succeeds but EXT4 fails

Error:
```PS
PS OsImageGenerator> go run .\tools\repro\ext4_panic_repro.go .\work\2025-10-01-raspios-bookworm-arm64-lite.img
Partition 1: success, fs.Type=0
Partition 2: PANIC: runtime error: integer divide by zero
goroutine 1 [running]:
runtime/debug.Stack()
        Go/src/runtime/debug/stack.go:26 +0x5e
main.tryFS.func1()
       OsImageGenerator/tools/repro/ext4_panic_repro.go:38 +0xad
panic({0x7ff7fd409400?, 0x7ff7fd5a9800?})
        Go/src/runtime/panic.go:783 +0x132
github.com/diskfs/go-diskfs/filesystem/ext4.groupDescriptorsFromBytes({0x7ff7fd605b80, 0x0, 0x0}, 0x0, 0x1fed6245, 0x2)
        go/pkg/mod/github.com/diskfs/go-diskfs@v1.5.0/filesystem/ext4/groupdescriptors.go:93 +0x2d4
github.com/diskfs/go-diskfs/filesystem/ext4.Read({0x7ff7fd48ae10, 0xc000008318}, 0x82000000, 0x20800000, 0xc000143da8?)
        go/pkg/mod/github.com/diskfs/go-diskfs@v1.5.0/filesystem/ext4/ext4.go:677 +0x3ed
github.com/diskfs/go-diskfs/disk.(*Disk).GetFilesystem(0xc00007a080, 0x2)
        go/pkg/mod/github.com/diskfs/go-diskfs@v1.5.0/disk/disk.go:239 +0x2f0
main.tryFS(0xc00000e1c0?, 0x2)
       OsImageGenerator/tools/repro/ext4_panic_repro.go:41 +0x59
main.main()
        OsImageGenerator/tools/repro/ext4_panic_repro.go:30 +0x1cb

Partition 3: error: unknown filesystem on partition 3
Partition 4: error: unknown filesystem on partition 4
```

Image has blank partitions at 3 and 4. This is normal

[...]

PR #435, opened 2026-09-18
fix: Fixed error when mkdiring on ext4 without 64-bit extension

lxgr-linux:
When creating directories on an ext45 without a 64-bit extension I got an `panic: runtime error: slice bounds out of range [:64] with capacity 32`.

This fixes the problem because the already sliced `checksumInput` introduced in https://github.com/diskfs/go-diskfs/pull/312 was sliced again and got an out of bounds.

Related to: https://github.com/diskfs/go-diskfs/issues/310


deitch (approved):
Nice catch, thank you.

deitch:
I see your email is on your profile; I will drop you a line, want to know what you are using it for. Besides, you have quite the array of interesting things!

